PR #2, opened 2017-08-06
Fix grammar

kossnocorp:
It's a great module and here is my humble contribution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Nano ID
 
-Very small secure URL-friendly unique ID generator.
+It's a very small and secure a URL-friendly unique ID generator.
 
 ```js
 var nanoid = require('nanoid')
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,20 +26,20 @@ Generator supports Node.js and [all browsers] from IE 11.
...
 ### Unpredictability
 
-Nano ID doesn’t use unsafe `Math.random()`. It use `crypto` module in Node.js
+Instead of unsafe `Math.random()` Nano ID uses `crypto` module in Node.js 
 and Web Crypto API in browsers.
 
 ### Uniformity
 
 `random % alphabet` is a popular mistake in ID generator. Change to get some
 symbols will be lower and it will reduce amount of guesses in bruteforcing.
```
kossnocorp: I've failed to parse `Change to get some symbols will be lower and it will reduce amount of guesses in bruteforcing`. It's worth to rewrite the sentence completely.

README.md
```diff
@@ -50,29 +50,29 @@ Nano ID use [better algorithm] and test uniformity:
...
 
 ```js
 var nanoid = require('nanoid')
 model.id = nanoid() //=> "Uakgb_J5m9g~0JDMbcJqLJ"
 ```
 
-Symbols `-,.()` are not encoded in URL, but in the end of link they could
-be detected as punctuation.
+Symbols such as `-,.()` aren't encoded in URL, but in the end of a link
+they could be identified as a punctuation symbol.
```
kossnocorp: I feel like the sentence isn't finished or lacks the conclusion why it matters that the symbols could be in interpreted as punctuation.

PR #7, opened 2017-08-09
Add TypeScript support

basarat:
🌹

### DIFF
--- a/format.d.ts
+++ b/format.d.ts
@@ -1,0 +1,2 @@
+declare const format: (random: (size: number) => string[], alphabet: string, size: number) => string;
+export = format;
--- a/generate.d.ts
+++ b/generate.d.ts
@@ -1,0 +1,2 @@
+declare const generate: (charset: string, length: number) => string;
+export = generate;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,2 @@
+declare const nanoid: () => string;
+export = nanoid;
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .DS_Store
+.alm
 *~
 
 node_modules/
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "browser": {
     "./random.js": "./random-browser.js"
   },
+  "types": "./index.d.ts",
   "devDependencies": {
     "docdash": "^0.4.0",
     "eslint": "^4.4.0",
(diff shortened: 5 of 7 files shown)


ai:
Great!

Can you also send types for `generate` and `format` functions?

basarat:
Done :rose:

ai:
@basarat Looks great! Few question:

1. What is `.alm` in `.gitignore`?
2. Do we need `tsconfig.json`?
3. Do we need to mention other `.d.ts` files in `package.json`?

basarat:
> What is .alm in .gitignore?

Just what alm uses http://alm.tools/ like `.idea` or `.vscode` etc. 

> Do we need tsconfig.json?

Helps make sure there isn't a TypeScript error in the `.d.ts` files 

> Do we need to mention other .d.ts files in package.json?

No. As `import * as generate from "nonoid/generate"` will automatically lookup a `generate.d.ts` file in the `node_modules/nanoid` folder :rose:

ai:
Great. Can you remove `.alm`?

If we have `tsconfig`, should we use it in `npm test`? Should we also add this config to `.npmignore`?

index.d.ts
```diff
@@ -0,0 +1,2 @@
+declare const nanoid: () => string;
```
BrainMaestro: `nanoid` takes a [number parameter](https://github.com/ai/nanoid/blob/ef1d664b52fb0cd978371c9601658f09c84c88cc/index.js#L21-L22)

generate.d.ts
```diff
@@ -0,0 +1,2 @@
+declare const generate: (charset: string, length: number) => string;
```
BrainMaestro: the actual [`generate`](https://github.com/ai/nanoid/blob/ef1d664b52fb0cd978371c9601658f09c84c88cc/generate.js#L21) functions's arguments are `alphabet` and `size`.

format.d.ts
```diff
@@ -0,0 +1,2 @@
+declare const format: (random: (size: number) => string[], alphabet: string, size: number) => string;
+export = format;
```
TrySound: Why not `export default`?

TrySound:
@basarat Will typescript find `index.d.ts` with `import nanoid from 'nanoid';`?

format.d.ts
TrySound: All modules here has only one export so you may don't use tricky funky synthetic exports.

ai:
Typings repositories sounds like good idea :) In this case only Flow/TS users will download type definitions
https://github.com/ai/nanoid/pull/15

PR #10, opened 2017-08-09
Replace true with action

marinintim:


marinintim:
Lint doesn't pass, though

ai:
It saves 2 bytes, but this optimization is not good code practice and could confuse person, who will read code :(

PR #14, opened 2017-08-09
Removed byte variable, replaced Math.ceil to binary OR, moved bytes var inside cycle, removed length lookups

mokhov:
-16bytes

### DIFF
--- a/format.js
+++ b/format.js
@@ -24,12 +24,12 @@
  * @name format
  */
 module.exports = function (random, alphabet, size) {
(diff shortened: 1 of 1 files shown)


marinintim:
It can skew distribution, skipping '0' as value.

ai:
I manually moved all changes that don’t reduce readability https://github.com/ai/nanoid/commit/40d0e426aec3ecbefc1ce586a3a660ba9e14e561

PR #15, opened 2017-08-10
Add flow types definitions

TrySound:


ai:
Can Flow use TypeScript type definitions? I don’t want to make npm package bigger by putting type definitions for every language :).

TrySound:
@ai Sadly no. And I don't think you need to choose. As you don't use type checkers accept both or send us to typings repositories.

ai:
Typings repositories sounds like good idea :) In this case only Flow/TS users will download type definitions

TrySound:
Yep

PR #38, opened 2017-10-25
Separate, size optimized implementation of index.js for browsers

subzey:
It turned out, `btoa`/`atob` are actually [parts of HTML5 spec](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa) now, and it's widely supported, so we could use it as a "shortcut" version for index.js in browsers.

The main idea is that the default dictionary shares 62 out of 64 chars with a base64 dictionary. And base64 preserves the uniform random distribution.

### DIFF
--- a/index-browser.js
+++ b/index-browser.js
@@ -1,0 +1,33 @@
+/* eslint-env browser, es5 */
+/**
+ * Generate secure URL-friendly unique ID.
+ *
+ * By default, ID will have 21 symbols to have a collision probability similar
+ * to UUID v4.
+ *
+ * @param {number} [size=21] The number of symbols in ID.
+ *
+ * @return {string} Random string.
+ *
+ * @example
+ * var nanoid = require('nanoid')
+ * model.id = nanoid() //=> "Uakgb_J5m9g~0JDMbcJqL"
+ *
+ * @name nanoid
+ */
+module.exports = function (size) {
+  size = size || 21
+  return (
+    btoa(
+      String.fromCharCode.apply(
+        0,
+        (window.crypto || window.msCrypto).getRandomValues(
+          new Uint8Array(-~size * 0.75)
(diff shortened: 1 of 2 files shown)


ai:
I will think about it after ReactiveConf.

ai:
Why we have `-~size * 0.75` and `.slice(0, size)`?

ai:
I made a simple benchmark and found that this way to convert random bytes to string is about 2 times slower (243 174 vs 411 223 ops/sec).

ai:
Sorry, performance is an important part of the project. We can reduce it dramatically for few bytes reducing. But the way was really smart!

subzey:
It's fine! Thank you!

> Why we have `-~size * 0.75` and `.slice(0, size)`?

It doesn't really matter now, but I won't leave it unanswered.

Base64 string is 4/3 (log<sub>64</sub>256) times longer than the source byte sequence. For example, to get 4 b64 chars we need 3 bytes. But that's the ideal case.

From 4 bytes we'll get a b64 string ranging from `AAAAAA==` to `/////w==`. Obviously, we can use only 1st 5 chars. The 6th gives non-uniform distribution and the 7th and 8th are garbage (b64 padding chars).

To compensate the effect of non-uniform distribution, we have to "overshoot" a little. `-~size * 0.75` is the same as `- (-1 + size | 0) * 0.75`, or `( (size | 0) + 1 ) * 3 / 4`. This value is then truncated down to the integer when used as an array length.

This way for the default size of 21 we get a b64 string ranging from
```AAAAAAAAAAAAAAAAAAAAAA==```
to 
```/////////////////////w==```

And after `.slice(0, 21)`:
```AAAAAAAAAAAAAAAAAAAAA```
to
```/////////////////////```

PR #39, opened 2017-10-30
176 b: Reverse loop, reuse var

subzey:
There's no difference, in which direction we're traversing the array of random values: they're random.

The loop was reversed, so the size can be reused as a loop counter.

I had to temporary disable a `yoda` eslint rule for one line. The reason is, Uglify cannot minify `i-- > 0` to `i-->0` because `-->` can be parsed as [a single token](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-html-like-comments).

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,8 +22,12 @@ module.exports = function (size) {
   size = size || 21
   var id = ''
   var bytes = random(size)
-  for (var i = 0; i < size; i++) {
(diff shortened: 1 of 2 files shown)


ai:
You are insane :D

ai:
Released in 1.0.1.

@subzey Can I ask you to write about it? I think my followers hate Nano ID tweets :D.

PR #51, opened 2018-02-23
Add cli

twhitbeck:
A couple times I've wanted to be able to generate a nanoid from the command line. This adds a cli to the package. If you're opposed to adding `minimist` as a dependency, I'd be willing to update the PR to parse the arguments manually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,27 @@ var url = require('nanoid/url')
 format(random, url, 10) //=> "93ce_Ltuub"
 ```
 
+### CLI
+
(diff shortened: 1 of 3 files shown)


ai:
Good idea. But we can afford dependency for this rate case. I have better idea. Do you want to create `nanoid-cli` package. In this case you can use good dependencies to parse arguments.

I will add a link to your project from docs.

What do you think?

twhitbeck:
excellent idea

ai:
Ping me when it will be ready or if you will need any help

PR #64, opened 2018-07-06
Make the chars order GZIP-friendly

subzey:
The string contains all the same characters: Alphanumeric and `~_`, but their order has changed:

For the code itself the order is not significant, but it's significant when serving a GZIPped code to the client. 

There's a very high chance words `crypto` and `getRandomValues` will be bundled somewhere near this string. And GZIP now has the opportunity to just copy the whole `getRandomV` and `cryp` string runs via backreferences (which of these "costs" only a few bits) instead of plain Huffman encoding.

Overall, this technique saves 8 .min.gz bytes - 5% of the entire entry size

### DIFF
--- a/.size-limit
+++ b/.size-limit
@@ -1,7 +1,7 @@
 [
   {
     "path": "index.js",
-    "limit": "162 B"
+    "limit": "154 B"
   },
   {
     "path": "generate.js",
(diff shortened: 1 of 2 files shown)


ai:
Awesome 😍

ai:
Released in 1.0.5

mxstbr:
🤯

PR #73, opened 2018-08-09
Use `|` operator instead of Math.floor()

oriSomething:
Use `|` operator instead of `Math.floor()`.

Now, my name *might* be saved in history ;-)

### DIFF
--- a/non-secure.js
+++ b/non-secure.js
@@ -22,7 +22,7 @@ module.exports = function (size) {
   size = size || 21
   var id = ''
   while (0 < size--) {
-    id += url[Math.floor(Math.random() * 63)]
+    id += url[Math.random() * 63 | 0]
   }
   return id
 }


ai:
Yeap https://github.com/ai/nanoid/commit/6781f482ac166ad1df529794097c8ed486adcfb2

Exact numbers depends on machine and Node.js version. But I saw changes in result after this commit. To be honest, I am not sure way. Maybe math operator works faster because it doesn't call a additional method?

oriSomething:
I think it's related to old optimizations for `asm.js`. `asm.js` uses a lot the `| 0` expression to cast number to integer so V8 added optimization that consider some expressions that ends with `| 0` as integers instead of floats. I'm not sure when `Math.floor` returned value treated as integer and when it just floored float, but `| 0` always become integer.
By the way the BuckleScript compiler also uses `| 0` optimization for intergers.

PR #78, opened 2018-08-27
Fix bug in non-secure.js

meiwhu:
The `Math.random` function returns a number in the range 0–1 (inclusive of 0, but not 1), so `Math.random() * 63 | 0` will never return 63.


ai:
I also improved tests to prevent regression 304c1b6

ai:
Released in 1.2.2

PR #84, opened 2018-09-27
Restore support for Node.js<v6.13.0

AndreyBelym:
Unfortunately I can't add an older Node config to the TravisCI env, because `eslint` requires `node>=6.14.0` (see the tests for the first commit)

### DIFF
--- a/async.js
+++ b/async.js
@@ -24,24 +24,28 @@ var random = require('./random')
 module.exports = function (size, callback) {
   size = size || 21
 
-  var randomPromise = random.async(size)
-    .then(function (bytes) {
-      var id = ''
-      while (0 < size--) {
-        id += url[bytes[size] & 63]
-      }
-      return id
-    })
-
   if (!callback) {
-    return randomPromise
+    return new Promise(function (resolve, reject) {
+      module.exports(size, function (error, id) {
+        if (error) {
+          reject(error)
+        } else {
+          resolve(id)
+        }
+      })
+    })
   }
 
-  randomPromise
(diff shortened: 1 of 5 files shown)


async.js
```diff
@@ -24,32 +23,25 @@ var url = require('./url')
...
-          resolve(id)
-        }
-      })
-    })
+
+  if (!callback) {
+    return randomPromise
   }
+
+  randomPromise
```
ai: It will not work :D. `callback` API was created to support environments without `Promise`.

random.js
```diff
@@ -1,8 +1,23 @@
...
+  return crypto.randomFillSync(Buffer.allocUnsafe(bytes))
+}
+
+module.exports.async = function (bytes) {
+  return new Promise(function (resolve, reject) {
+    function callback (error, buffer) {
+      error ? reject(error) : resolve(buffer)
+    }
+
+    if (crypto.randomFill) {
```
ai: If will be better to:

```js
if (crypto.randomFill) {
  module.exports.async = …
} else {
  module.exports.async = …
}
```

Will reduce extra check in ``random`.

ai: Yeap, I know that it is harder to test it :D

async.js
AndreyBelym: Ok, I've got it. So I can't use Promise in `random.js` too :(

AndreyBelym:
FPR

ai:
Released in 1.2.5. Thanks for the help.

PR #91, opened 2018-10-07
Reduce url symbols size for 3 bytes

philosaf:
Some symbols movement for improving gzip size from 73 bytes to 70 (checked with size-limit)

### DIFF
--- a/url.js
+++ b/url.js
@@ -1,6 +1,9 @@
 /**
  * URL safe symbols.
  *
+ * This alphabet uses a-zA-Z0-9_~ symbols.
+ * (Symbols order was changed for better gzip compression.)
+ *
  * @name url
  * @type {string}
  *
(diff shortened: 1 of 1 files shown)


ai:
😄 sure. Let's just explain alphabet and trick in the comment.

Compare to `index.js` tricks are less expected here.

ai:
Also, maybe you want to implement this task, to be sure that it is the best alphabet order

https://cultofmartians.com/tasks/nanoid-brute.html

ai:
Yeap, I know how it works.

But when users will try to find what alphabet is used in Nano ID, user will be confused about `url.js` content. This is why we need to explain it in commend in the code (like “this alphabet uses `a-zA-Z0-9_~`, we changed symbols order for better gzip compression”).

philosaf:
Done.
Did not have time to fix `yaspeller` errors :smile:

ai:
Released in 1.3

PR #93, opened 2018-10-10
Snippet to demonstrate async RNG in ReactNative; Fixes #88

### LINKED ISSUE #88: Can I help?  Is it possible to generate unguessable nanoid's on ReactNative?
Hey,

Thanks for building this - very cool and I appreciate all the hard work you do on it!  :-)

I notice that the README says nanoid doesn't generate unpredictable ids on RN due to the lack of a secure RNG.  However, would it be possible to use something like https://www.npmjs.com/package/react-native-securerandom for example?

I was thinking to avoid a dependency on this third-party RN package, maybe we could use dependency injection to inject a secure RNG, user's choice, and give instructions for simultaneously adding this `react-native-securerandom`.  So current users would not be impacted since current defaults would work.  But new users could pass in a reference to some function that uses `react-native-securerandom` to supply the secure random bytes (via a promise).

I'd be happy to work on a PR for this if there's interest, but could you give me a few pointers where to get started in nanoid, like where to do the dependency injection and roughly how and I could give it a shot?

Thanks1

joelgetaction:
Fixes #88

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,33 +15,30 @@ It uses [Size Limit] to control size.
 As result it could reduce ID size from 36 to 21 symbols.
 
 ```js
-var nanoid = require('nanoid')
-model.id = nanoid() //=> "V1StGXR8_Z5jdHi6B~myT"
+var nanoid = require("nanoid");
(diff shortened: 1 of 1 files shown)


ai:
Released in 1.3

PR #101, opened 2018-11-06
Update README.md

DaniilDem:
Use warning for React


ai:
Good idea. But it is very easy to misread it and think that it is a recommendation. I suggest:

1. Make `Do not` as `<strong>`.
2. Show how to do example (with nanoid in comments)

PR #116, opened 2019-03-25
test: add benchmark for secure-random-string

markstos:
secure-random-string has comparable performance for the async API despite
returning 32 characters by default, but is slower in the sync benchmark.

I tried benchmarking returning on 21 characters for a more comparable test, but
it was hardly faster.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "parcel-bundler": "^1.12.1",
     "rimraf": "^2.6.3",
     "rndm": "^1.2.0",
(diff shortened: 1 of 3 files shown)


ai:
Wait for a second, I will restart my laptop to have a clean environment for the benchmark.

ai:
Here is the result after running a benchmark (just after the system was restarted, without any GUI program)

https://i.imgur.com/zRWXwMA.jpg

2f757ba

markstos:
Those are similar to my results.

PR #117, opened 2019-03-26
es modules for modern browsers

chase-moskal:
good day!

i wanted to load nanoid in a modern browser the hip new way, so i made this experimental pull request and published `nanoid-esm@2.1.0-esm.1`

nanoid-esm is available now on unpkg: https://unpkg.com/nanoid-esm@2.1.0-esm.1/dist/index.browser.js

[example on codepen](https://codepen.io/ChaseMoskal/pen/XGQBBy?editors=1000)

this pull request introduces babel compilation to produce es modules in `dist/`

the commonjs modules are now available in `dist/cjs/`

source code is organized in `source/`

of course this approach is highly biased towards es modules, demoting the commonjs to a second class citizen in `dist/cjs/`, and i had to change all of the `require` paths in the readme examples... this project might want to flip it, such that the es modules are located in a `dist/esm/`...

also, the es modules are currently ending in the extension `.js` instead of the experimental `.mjs` because i can tell which way the wind is blowing -- and anyways, this pr focuses on browser support -- node just doesn't quite yet support esm like browsers do


chase-moskal:
**this pr breaks the build.**
1.  lint fails on `source/test/alphabet-worker.js` "worker_threads" is not found  node/no-missing-require, not sure why this popped up
2. size-limit fails, not sure why
    ```
    source/index.js
    Package size limit has exceeded by 95.47 KB
    Package size: 95.61 KB
    Size limit:   141 B
    ```

ai:
> size-limit fails, not sure why 

Because you increase the size of Nano ID from 141 bytes to 95.61 KB.

If you want to use Babel, you need to be careful and be sure what it generates.

ai:
> i wanted to load nanoid in a modern browser the hip new way, so i made this experimental pull request and published nanoid-esm@2.1.0-esm.1

What is the exacty use case for ES modules?

You suggest making the build process much more complicated. We need a really good reason to do it.

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 2.1.0-esm.0
```
ai: Can you remove it? In my project changelog changes are part of release commit

README.md
```diff
@@ -156,8 +156,8 @@ To generate secure random IDs in React Native, you must use
 
 ```js
 const generateSecureRandom = require('react-native-securerandom').generateSecureRandom
-const format = require('nanoid/async/format')
-const url = require('nanoid/url')
+const format = require('nanoid/dist/cjs/async/format')
+const url = require('nanoid/dist/cjs/url')
```
ai: It is a breaking change. It can’t be done in 2.1 and you need a really good reason to make 3.0.

README.md
```diff
@@ -190,10 +190,27 @@ If you don’t need unpredictable IDs, but you need Web Workers support,
...
+
+es modules are available under `nanoid/dist/`
+
+commonjs modules are available under `nanoid/dist/cjs/`
+
+[codepen example](https://codepen.io/ChaseMoskal/pen/XGQBBy?editors=1000)
+
+```html
+<script type="module">
+  import nanoid from "https://unpkg.com/nanoid-esm@2.1.0-esm.1/dist/index.browser.js"
```
ai: Using `unpkg.com` is a bad practice since it will add DNS request, TCP connection and one more internet latency to the website loading process. It will slow down the website up to 0.5s.

1. Browser loads `example.com`
2. DNS request, TCP connection, Internet latency
3. Browser see `example.com/index.js` link and make new request.
4. Internet latency
5. Browser see `https://unpkg.com/nanoid-esm@2.1.0-esm.1/dist/index.browser.js` link
6. **DNS request, TCP connection, Internet latency**

With npm and Parcel/webpack you will not have this step 6.

package.json
```diff
@@ -60,31 +71,31 @@
   },
   "size-limit": [
     {
-      "path": "index.js",
+      "path": "source/index.js",
```
ai: You need to use build file here to check what is the size after Babel.

package.json
```diff
@@ -108,12 +119,14 @@
     "*.md": "yaspeller-ci",
     "*.js": "eslint"
   },
-  "husky": {
+  "husky---DISABLED": {
```
ai: Why did you disable Husky? If you want to push some broken code to discuss it, you can add `--no-verify` argument to `git commit -m "message" --no-verify` command.

CHANGELOG.md
chase-moskal: that makes sense

i don't have any time soon

README.md
chase-moskal: yes of course

i think, if nanoid was ever to provide esm format, it would be best to abandon this pr, and take it from a better approach:

1. leave all the commonjs files in-place (don't organize them into a `source` dir)
2. compile esm to `dist-esm`

this way, esm could be a non-breaking addition, which would be much better

README.md
chase-moskal: an excellent point of course

i think tools like unpkg and jsdelivr are best used during development

optimization options include
- using rollup or similar to create an optimized tree-shaken bundle
- using a lot of [`<link rel="modulepreload" href="..."/>`](https://www.chromestatus.com/feature/5762805915451392) to give the browser a heads-up

package.json
chase-moskal: doh, good point, these should be pointing at

`dist/cjs/index.js`

or

`dist/index.js`

not in `source`

package.json
chase-moskal: `--no-verify` — good tip, i didn't know!

i haven't used husky before, i was just rashly trying to get my changes committed and onto my github -- even if they are bad! ;)

you see, i have constant paranoia that my hard drive will crash at any moment, and so i get very antsy and sweaty when i have staged changes for more than 15 minutes :D

chase-moskal:
> Because you increase the size of Nano ID from 141 bytes to 95.61 KB.

ah interesting, i suppose it's measuring the size of the whole package, not the individual files.. of course if there are two module formats (es and cjs) then a consumer would only be using half of the package..

hey at least i added some spellcheck keywords -- you see, *some* effort was legitimately made :D

> What is the exacty use case for ES modules?

i'm experimenting on a project that exclusively uses the new module system, and the use-case is simply that i'd like to use nanoid in the browser :)

so i'm using the `<script type="module"></script>` workflow. it's not compatible with commonjs, so i expect that in the not-so-distant future (particularly once node finally supports esm properly), package authors might start switching en masse to ESM as the primary format, then using CJS as a fallback for legacy consumers

it's interesting that browsers are so far ahead of node on this topic.. i suppose competition in the browser wars might provide an extra incentive compared to node... then again, [deno](https://deno.land/) knows what's up on this topic :)

i should have made more clear from the start: that this PR is the (potentially throwaway) result of me staying up all night and disjointedly exploring es modules -- i anticipate that if this PR has any positive value, it might be simply discussing a *better way* to implement es modules than the approach in this PR -- hopefully you don't feel that i've merely wasted your time ;)

ai:
> ah interesting, i suppose it's measuring the size of the whole package, not the individual files

Nope. [Size Limit](https://github.com/ai/size-limit/) is using to calculate the real cost for user’s JS bundle size of adding `require(your-lib)`.

It creates an empty webpack project, adds your library to it and then gets the difference between before and after client-side JS bundle size.

> so i expect that in the not-so-distant future (particularly once node finally supports esm properly), package authors might start switching en masse to ESM as the primary format, then using CJS as a fallback for legacy consumers

Yeap, I am thinking about it. But I am still waiting for the final modules spec from npm and how ES modules bundlers will work.

I am afraid that switching to modules right now could hurt users more, that bringing new things.

It is OK for some experimental project, but not so good for big projects like Nano ID.

> it's interesting that browsers are so far ahead of node on this topic

I am not happy that they are so ahead. There is no solution for latency problem (we can fix it by a big list of `preload` links, but it will bloat HTML size on the real projects with 100+ files).

Maybe we need some new generation of bundlers which will concat similar files to the module (to keep the number of the modules around 10-50) and generate `preload` list.

> that this PR is the (potentially throwaway) result of me staying up all night and disjointedly exploring es modules

If you want you can keep ESM version in `nanoid-es` and you can add link to the Tools section of Nano ID docs.

chase-moskal:
i've created a new branch, [esm2](https://github.com/chase-moskal/nanoid/tree/esm2), which is much better

after i have a little help figuring out #118, i'll fix c105ce2 and post a new PR

chase-moskal:
new pull request #119 is much better :+1:

PR #126, opened 2019-05-27
Update README.md

FabianoLothor:
Correction.


ai:
You do not need `"input"` prefix for `key`. You need it only for HTML ID.

FabianoLothor:
I know, but the text is not correct with the code.

> This is good code. this.id will be generated only once. Note, that we added "input" string in front of id, because Nano ID could be started from number. HTML ID can’t be started from the number.

ai:
Oops. I fixed the text 5330938

PR #132, opened 2019-07-18
Adds documentation for react-native using expo

benvp:
I added some documentation how to use nanoid with react-native when using expo. Had to search a while to find `expo-random` and to get nanoid up and running, so I thought it would be worth sharing in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,22 +185,7 @@ Nano ID could be started from number. HTML ID can’t be started from the numbe
 
 ### React Native
 
-To generate secure random IDs in React Native, you must use
-[a native random generator] and asynchronous API:
-
-```js
-const generateSecureRandom = require('react-native-securerandom').generateSecureRandom
-const format = require('nanoid/async/format')
-const url = require('nanoid/url')
-
-async function createUser () {
-  user.id = await format(generateSecureRandom, url, 21);
-}
-```
-
-#### Expo
-
-If you are using expo, you can not use [a native random generator] because it requires native-linking. Anyway you can use [expo-random] as the secure number generator.
+To generate secure random IDs in React Native, you must use a native random generator (like [expo-random] or [react-native-securerandom]) and the asynchronous API:
 
 ```ts
 import { getRandomBytesAsync } from 'expo-random';
@@ -213,10 +198,8 @@ const random = (getRandomBytesAsync as unknown) as (byteCount: number) => Promis
 export const id = async () => format(random, url, 21);
 ```
 
-Also, if you don't need secure id's there is still the non-secure version, which does not require a secure number generator (only use if you don't care about predictable IDs).
-
 [expo-random]: https://github.com/expo/expo/tree/master/packages/expo-random
-[a native random generator]: https://github.com/rh389/react-native-securerandom
+[react-native-securerandom]: https://github.com/rh389/react-native-securerandom
 
 
 ### PouchDB and CouchDB
--- a/README.md
+++ b/README.md
@@ -189,12 +189,14 @@ To generate secure random IDs in React Native, you must use a native random
 generator (like [expo-random] or [react-native-securerandom]) and the
 asynchronous API:
 
-```ts
-import { getRandomBytesAsync } from 'expo-random';
-import format from 'nanoid/async/format';
-import url from 'nanoid/url';
+```js
+const getRandomBytesAsync = require('expo-random').getRandomBytesAsync
+const format = require('nanoid/async/format')
+const url = require('nanoid/url')
 
-export const id = () => format(getRandomBytesAsync, url, 21);
+async function createUser () {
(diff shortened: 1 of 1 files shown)


ai:
Is this random generator better than `react-native-securerandom`?

benvp:
Honestly, I have no idea. It's just the only generator I found which works together with expo. The `expo-random` description says:

> Provides a native interface for creating strong random bytes. With Random you can create values equivalent to Node.js core crypto.randomBytes API.

So I would guess it's like the NodeJS crypto implementation.

ai:
What is Expo and why it is a big deal for React Native?

I am afraid that the example with any RN random library will make docs more complicated. We can list all generators in docs with only one example.

benvp:
Just had a quick look into the source of `react-native-securerandom` and `expo-random` and it looks like they are using almost the same implementation for each android and ios.

I understand that having an example for any random library doesn't make sense. In case of expo, it's a bit different. Expo originated from `create-react-native-app` and is now a broad eco-system which allows the development of react-native apps without touching any native code. As the native codebase is abstracted, it is currently not possible to integrate libraries which have native modules. For example `react-native-securerandom` requires linking native modules. That is not possible when using the expo ecosystem. 

I would argue that expo is quite a big thing and not a random-library. It has a huge adoption and is the recommended way to start new react-native projects  (see https://facebook.github.io/react-native/docs/getting-started). At least in my opinion, it would make sense to mention it in the readme - especially because there is already a section covering react-native.

ai:
Let’s replace `react-native-securerandom` to `expo-random` in the example and add link to `react-native-securerandom` in the bottom. Can you update your PR?

README.md
```diff
@@ -185,20 +185,21 @@ Nano ID could be started from number. HTML ID can’t be started from the numbe
 
 ### React Native
 
-To generate secure random IDs in React Native, you must use
-[a native random generator] and asynchronous API:
+To generate secure random IDs in React Native, you must use a native random generator (like [expo-random] or [react-native-securerandom]) and the asynchronous API:
```
ai: All docs use 80 columns limit. You need to respect this rule too.

benvp:
@ai I updated the PR. Please have a look if it's okay for you.

README.md
```diff
@@ -185,20 +185,21 @@ Nano ID could be started from number. HTML ID can’t be started from the numbe
...
+```ts
+import { getRandomBytesAsync } from 'expo-random';
+import format from 'nanoid/async/format';
+import url from 'nanoid/url';
 
-async function createUser () {
-  user.id = await format(generateSecureRandom, url, 21);
-}
+// Only required if you use TypeScript to tell the compiler that the types are compatible.
+const random = (getRandomBytesAsync as unknown) as (byteCount: number) => Promise<number[]>;
```
ai: I use pure JS everywhere. Let’s use pure JS here too. TS users will find a way to do it.

README.md
```diff
@@ -185,20 +185,21 @@ Nano ID could be started from number. HTML ID can’t be started from the numbe
...
+
+export const id = async () => format(random, url, 21);
```
ai: Can we use top-level `async` in all environments right now?

README.md
```diff
@@ -185,20 +185,20 @@ Nano ID could be started from number. HTML ID can’t be started from the numbe
...
+asynchronous API:
 
-```js
-const generateSecureRandom = require('react-native-securerandom').generateSecureRandom
-const format = require('nanoid/async/format')
-const url = require('nanoid/url')
+```ts
+import { getRandomBytesAsync } from 'expo-random';
+import format from 'nanoid/async/format';
+import url from 'nanoid/url';
```
ai: All examples have no semicolons. It will be strange for users to see different style guides in one doc.

README.md
```diff
@@ -185,20 +185,20 @@ Nano ID could be started from number. HTML ID can’t be started from the numbe
...
 
-async function createUser () {
-  user.id = await format(generateSecureRandom, url, 21);
-}
+export const id = () => format(getRandomBytesAsync, url, 21);
```
ai: Should we rename `id` to `getId`? But we need to clear that value will be wrapped to Promise. This is why have `createUser` code in the previous example to explain how to use the async code here. Do you have reasons to avoid this example?

README.md
ai: Also, note that your ES import syntax is incorrect. ES imports require a full path. This imports will work only in Babel. It was a reason why I used `require` in all other docs. If you want to change `require` to `import` you need to do it across all docs (and first start with good reason of doing it)

ai:
Thanks for all small changes. Sorry for long lost of fixes.

benvp:
No worries, they make sense :)

ai:
Oh. The last change. Can you put library names as a code in text to fix spelling check error on CI (spelling check avoids code).

PR #136, opened 2019-08-12
Allow react-native alternative for a async random number generator

swernerx:
My colleagues are building a react-native application. We use `nanoid` as part of our login flow. While introducing our common solution I got an issues reported that the current usage of `nanoid` breaks the application for them. The reason for this is the exception which is being thrown in `index.browser.js`. (Interestingly the same exception should be probably added to the async layer - which is not yet implemented.) Back to the issue. We could probably think of adding [a `react-native` entry](https://github.com/facebook/metro/blob/a29d30327365f3f52652f68d53896355021cc693/packages/metro/src/node-haste/Package.js#L45) as supported by the Metro bundler (used by React Native) to use a special random number generator in async. 

For our login flow I am fine with using the async API. But still this has to work without throwing errors in React Native. 

The solution described in the Readme is a good solution for plain react-native projects, but does not correctly work in these cross-target libraries. I do not want to make any native code for iOS/Android a dependency to our pure JS library. 

So I wondered how this can be probably look like for `nanoid` if supported out of the box. I just replaced the random number generator with the code listed in the readme. Do you think it would be that simple? 

This PR is meant as a suggestions and discussion starting point. Thanks for listening.

### DIFF
--- a/async/random.reactnative.js
+++ b/async/random.reactnative.js
@@ -1,0 +1,18 @@
+var expoRandom
+
+try {
+  /* eslint-disable-next-line global-require, node/no-missing-require */
+  expoRandom = require('expo-random')
+} catch (importError) {}
+
+module.exports = function (bytes) {
+  if (!expoRandom) {
+    throw new Error(
+      'React-Native does not have a built-in secure random generator. ' +
+      'Install "expo-random" locally or ' +
+      'if you don’t need unpredictable IDs, you can use `nanoid/non-secure`.'
+    )
+  }
+
+  return expoRandom.getRandomBytesAsync(new Uint8Array(bytes))
+}
--- a/package.json
+++ b/package.json
@@ -24,6 +24,9 @@
     "./async/index.js": "./async/index.browser.js",
     "./async/random.js": "./async/random.browser.js"
   },
+  "react-native": {
+    "./async/random.js": "./async/random.reactnative.js"
+  },
   "sideEffects": false,
   "devDependencies": {
     "@logux/eslint-config": "^30.0.2",


ai:
Wow, I didn’t know about `react-native`. Thanks for the PR.

ai:
> So I wondered how this can be probably look like for nanoid if supported out of the box. I just replaced the random number generator with the code listed in the readme. Do you think it would be that simple?

I do not know a simple solution. You suggested the best solution from what I know right now:

1. Using `nanoid/non-secure` if it is not important
2. Or write a custom wrap-around Nano ID with using some hardware random generator library for RN

swernerx:
Mhh, do you have a react-native specific test somewhere? I will ask my colleagues tomorrow to check it out, but it would be probably helpful for regression checks.

swernerx:
Thanks for merging it that fast BTW.

ai:
> Mhh, do you have a react-native specific test somewhere?

Nope. Can I ask you to test this `react-native` key in `package.json`? Just take Nano ID from GitHub: `"nanoid": "ai/nanoid"`

swernerx:
I'll do.

ai:
Another way to show a warning for RN users is to use development only checks:

https://github.com/ai/nanoid/blob/master/index.browser.js#L1-L8

Does it work in RN right now?

swernerx:
You didn't change something there, right? The throw is crashing for my collegues... I figure they are in development mode as well while developing the app.

ai:
Nope, [this error](https://github.com/ai/nanoid/blob/master/index.browser.js#L1-L8) should be visible for your colleagues if they are using development mode and I checking Web Crypto API support correctly

swernerx:
Right now they see this exact error. This prevent booting the application. Probably it would be more sensible to warn about it and use a non-secure alternative instead (transparently).

swernerx:
Alternatively moving the throwing into the actual implementation code would be better in a sideEffects-free world. Right now it throws just because it is somehow imported, but not even used directly.

ai:
The warning can be missed and then they will have problems in production. It was an error for the reason because loading `nanoid/index` in RN is a mistake (it can’t work correctly, because this method can’t get another random generator)

ai:
> Right now it throws just because it is somehow imported, but not even used directly.

I can move error into the function. But I think that some developers may miss it during the quick look. I think importing is the issue itself.

ai:
Any updates?

swernerx:
Sorry, seems by colleagues are busy with other stuff right now.

ai:
Since I didn’t find any mentions of `react-native` key, I reverted changes but used your test in the better error message for React Native.

If `react-native` actually works we can return your code later.

But thanks for the PR, since we now should have much better DX or React Native.

swernerx:
The 'react-native' key is supported by the metro bundler. I even shared the source code to the place where it is implemented. When existing it is preferred over the other entries e.g. `main`

https://github.com/facebook/metro/blob/a29d30327365f3f52652f68d53896355021cc693/packages/metro/src/node-haste/Package.js#L45

It also supports `resolverMainFields` like Webpacks `mainFields` to configure this. What else do you required for keeping the code?

swernerx:
Yes. Indeed!

ai:
Better RN released in 2.1

PR #141, opened 2019-09-10
Fix react native usage

shawnhwei:
Some small fixes for react native.

Maybe its a Windows thing, but the path for the `random.rn.js` in `package.json` should be relative to the root.

> Unable to resolve "./random" from "node_modules\nanoid\async\index.js"

Also, the call to `expo-random` was a bit off.

> [Unhandled promise rejection: TypeError: expo-random: getRandomBytesAsync(0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0) expected a valid number from range 0...1024]

Correct method signature is: `getRandomBytesAsync(byteCount: number): Promise<Uint8Array>`

### DIFF
--- a/async/random.rn.js
+++ b/async/random.rn.js
@@ -10,5 +10,5 @@ try {
 }
 
 module.exports = function (bytes) {
-  return random.getRandomBytesAsync(new Uint8Array(bytes))
+  return random.getRandomBytesAsync(bytes)
 }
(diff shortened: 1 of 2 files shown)


ai:
Thanks. Released in 2.1.1.

Do you have Twitter account to mention you in release tweet?

shawnhwei:
I do not. Thanks!

PR #146, opened 2019-10-15
use .bind to reduce bundle 2b more

Jabher:
I'm unsure about whether you want to use .bind - but from my point of view it is pretty usable (IE9+, even more than getRandomValues reference) and somehow exactly `null` (neither `undefined` nor `0` or whatever) produces 2kb less.

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -19,6 +19,4 @@ var format = require('./format')
  * @name generate
  * @function
  */
-module.exports = function (alphabet, size) {
-  return format(random, alphabet, size)
-}
+module.exports = format.bind(null, random)


TrySound:
kb -> b

Jabher:
🤦
totally lost the habit of measuring something in bytes

Jabher:
@ai note - @TrySound pointed on commit message, it is slightly wrong (2kb, should be 2b).
I was planning to do it later, but you already merged, so probably you should reword it on your own now

ai:
Released in 2.1.4

PR #149, opened 2019-10-16
chore: remove microtime dependency

CyberAP:
Microtime is not used anywhere and breaks project scaffolding in Windows.


ai:
Thanks

PR #150, opened 2019-10-16
chore: add missing size-limit dependency

CyberAP:


ai:
`@size-limit/preset-small-lib` already installs `size-limit`. Do you have some problem with this method?

CyberAP:
Yes, when running `npm run test` I get an error that size-limit is an unkown command.

ai:
We do not have such problem on Travis CI. Is it Windows specific problem?

Can you try to replace `size-limit` in `scripts` to.`npx size-limit`?

CyberAP:
I think I had a corrupt package-lock.json. Making a clean install fixed the issue. Sorry for the inconvenience.

PR #151, opened 2019-10-17
golf: reduce generate size by 3 bytes

CyberAP:
- Hoist id declaration, allowing for var concatenation from minifier

- Increment index in-place


CyberAP:
Turns out I was late 😂

PR #155, opened 2019-10-23
golf: Remove extra branching, reduce generate size by 3 bytes

CyberAP:


CyberAP:
I am not 100% sure declaring `i` and `bytes` in `for` loop wouldn't break the code, but tests prove me wrong.

PR #157, opened 2019-11-14
Reduce url.js size using genetic algorithm

polemius:
I have also written the [blog post](https://dev.to/polemius/how-i-reduce-the-size-of-library-with-genetic-algorithm-6m8) about this PR.

### DIFF
--- a/async/index.browser.js
+++ b/async/index.browser.js
@@ -4,7 +4,7 @@ var crypto = self.crypto || self.msCrypto
  * This alphabet uses a-z A-Z 0-9 _- symbols.
  * Symbols order was changed for better gzip compression.
  */
-var url = 'Uint8ArModuleSymbhasOw-012345679BCDEFGHIJKLNPQRTVWXYZ_cfgjkpqvxz'
+var url = 'IUint8Ar21ModulvezGFYPCJ7_p0V4XSymbLBNH6fTqQ35xD9ZREghasOw-cjkWK'
 
 module.exports = function (size) {
   size = size || 21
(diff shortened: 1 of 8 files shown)


ai:
Oh, wow! :D

1. Can you add Size Limit changes to the PR?
2. Can we remove `test/alphabet`? Or maybe we should rename `test/genetic-algorithm.js` to `test/alphabet`?
3. Can we reduce also other `*.browsser.js` files to see how method works in different files?

polemius:
1. I've changed the size limit option for url.js. Did I miss something?
2. Yes, sure. I've changed it
3. I will run this algorithm with `index.browser.js` and we will see

ai:
> I will run this algorithm with index.browser.js and we will see

Can you run it for `async/index.brower.js`, `random.browser.js` and `url.js`?

> Yes, sure. I've changed it

Hm, maybe I was wrong. Seems like the best strategy is to use old `test/alphabet` and then GA one. It will get the best result. Sorry for asking,  but can you revert `test/alphabet` removing (if I am right that they work better together).

polemius:
I've reduced the size of `index.js` by 2 bytes 🎉  and `non-secure/index.js` by 1 byte. It is so cool!

ai:
It’s really put Nano ID to next level of how much computer science can be used for the simple lib :D

ai:
Released in 2.1.7.

I will post tweets during next week.

polemius:
I've leaved this algorithm for all day and still no new results. I'm just thinking about changing also the variables in the code. Maybe the algorithm can find solution where total length of string will be higher but gziped size lower 🤔 
But again it was very fun and cool! Thank you for [this comment](https://github.com/ai/nanoid/pull/94#issuecomment-428763984). Because of it I made this PR :)

PR #159, opened 2019-12-14
Adding example for react using hooks

nahumzs:
This helps to understand how to use nanoid with react hooks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,13 +186,10 @@ This is good code. `this.id` will be generated only once:
 }
 ```
 
-or
+or with hooks
 
 ```jsx
-function Element() {
-   const id = React.useRef(nanoid()).current;
-   return <Item key={id} />
-} 
+const [id] = React.useState(nanoid)
 ```
 
 If you want to use Nano ID for `id`, you must to set some string prefix.
--- a/README.md
+++ b/README.md
@@ -186,6 +186,15 @@ This is good code. `this.id` will be generated only once:
 }
 ```
 
+or
+
+```jsx
+function Element() {
+   const id = React.useRef(nanoid()).current;
+   return <Item key={id} />
+} 
+```
+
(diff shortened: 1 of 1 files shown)


ai:
Am I right, that `nanoid()` will be called on every component render? I could be bad for performance.

What do you think aboiut this way to use Nano ID with hooks?

```
const key = useMemo(nanoid)
```

TrySound:
@ai in this case nanoid is called on every render too unless you specify dependencies list. And still useMemo is not stable. It works only for pure computations. React may drop cache any time. These are ways to go.

```js
const useNanoid = () => {
  const ref = React.useRef(null)
  if (ref.current == null) {
    ref.current = nanoid();
  }
  return ref.current;
}

const Component = () => {
  const id = useNanoid()
}
```

```js
const Component = () => {
  const [id] = React.useState(() => nanoid());
}
```

nahumzs:
I put a small example on place https://codesandbox.io/s/nanoid-with-hooks-yv1u5 ref

TrySound:
Did you save it? I don't see any nanoid usage.

nahumzs:
@TrySound sorry my bad is there now.

ai:
@nahumzs OK, seems like `useMemo` doesn't work. What do you think about `const [id] = useState(nanoid)`?

TrySound:
`useMemo(() => nanoid(), [])` does work but eventually may break because nanoid is not pure computation.

nahumzs:
`const idState = React.useState(() => nanoid())` this will work as well and is a smoother syntax than ref (I updated the example)

ai:
ref has a problem with calling Nano ID on every component render. It didn't change the ID, but use CPU.

ai:
Thanks for improving docs :)

PR #161, opened 2020-01-07
Replace Math.ceil with the bit operation

aidarkhanov:
Shaved off some bytes 🎉
Tests run successfully and no performance regression was detected.

How it works:
```js
-~2.71 === 3
Math.ceil(2.71) === 3
```

Pitfall (ceiling integers):
```js
-~2 === 3 // returns true
Math.ceil(2) === 3 // returns false
```

### DIFF
--- a/async/format.browser.js
+++ b/async/format.browser.js
@@ -1,6 +1,6 @@
 module.exports = function (random, alphabet, size) {
   var mask = (2 << 31 - Math.clz32((alphabet.length - 1) | 1)) - 1
-  var step = Math.ceil(1.6 * mask * size / alphabet.length)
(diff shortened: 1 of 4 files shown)


ai:
Good idea. But let's avoid jack from non-browser version like in `format.js` (without `.browser` in the name).

aidarkhanov:
Do you mean we shall remain Math.ceil in format.js and async/format.js files?

ai:
Yeap

aidarkhanov:
Ok, anything else?

ai:
What is your Twitter to mention you in release announce?

aidarkhanov:
@dairaidarkhanov

ai:
Thanks. Released in 2.1.9.

Issue #162, opened 2020-01-07
ReactNative claims that expo-random is not installed in dev mode.


skaldo:
When you run RN app in dev mode with attached debugger, index.browser.js gets evaluated. An exception is then thrown that expo-random is not installed.

This is due to the platform-only check, in index.browser we do not check for the presence of expo random. This check is done in `async/random.rn.js`, thus check in `index.browser.js` is imo redundant.

I do propose to remove the check from index.browser.js. This itself resolves the issue, but referencing this file from RN runtime is not good, thus I would recommend to alter package.json to provide correct index.js.

<img src="https://user-images.githubusercontent.com/396663/71902080-9afbbf80-3161-11ea-8b13-0ae0a69f8799.png" height="500">

ai:
Why plan from error text (of using non-secure or `nanoid/async`) doesn't work in your case?

skaldo:
Of course I do have expo-random installed and it works, but when you are in debug mode (chrome debugger attached) the index.browser.js is executed. Here we do not check for expo-random, but just throw an exception that prevents the RN application from being executed.

Fundamental issue is, that the `index.browser.js` is executed at all - this is fixed by my change in package.json.
Second issue is the check itself in index.browser.js, which is redundant with the one in `async/random.rn.js`.
My PR takes care of both things.

skaldo:
Thanks to the help of @ai I found that there was a dependency conditionally requiring `nanoid` or `nanoid/non-secure` based on the existence of `crypto`.

An effect of this was the fact, that this library was throwing an exception in the debug mode only.

Anybody with similar problem should check the comments in the linked PR.

PR #163, opened 2020-01-07
Fix: secure-random-number-generator not found in RN debug mode.

skaldo:
https://github.com/ai/nanoid/issues/162

### DIFF
--- a/index.browser.js
+++ b/index.browser.js
@@ -1,11 +1,4 @@
 if (process.env.NODE_ENV !== 'production') {
-  if (typeof navigator !== 'undefined' && navigator.product === 'ReactNative') {
-    throw new Error(
-      'React Native does not have a built-in secure random generator. ' +
-      'If you don’t need unpredictable IDs, you can use `nanoid/non-secure`. ' +
-      'For secure ID install `expo-random` locally and use `nanoid/async`.'
-    )
-  }
   if (typeof self === 'undefined' || (!self.crypto && !self.msCrypto)) {
     throw new Error(
       'Your browser does not have secure random generator. ' +
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "./async/random.js": "./async/random.browser.js"
   },
   "react-native": {
+    "./index.js": "./index.js",
     "./async/random.js": "./async/random.rn.js"
   },
   "sideEffects": false,
--- a/test/react-native.test.js
+++ b/test/react-native.test.js
@@ -2,8 +2,8 @@ global.navigator = {
   product: 'ReactNative'
 }
 
-it('warns to use non-secure generator on old browsers', () => {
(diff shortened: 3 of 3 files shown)


ai:
In this PR you force RN to use Node.js version. Node.js version use `crypto` Node.js module.

Is this module now available in RN?

skaldo:
RN uses the `index.js` that then imports `./random.rn.js` as specified in package.json for RN environment.
`random.rn.js` does use `expo-random` and not the `crypto` module.

ai:
@skaldo Yeap, we did it because RN doesn't support `crypto`.

Did it got `crypto` support recently?

skaldo:
Your implementation in `random.rn.js` is pretty correct. Just the path how to get to this file is not.

> RN uses the index.js that then imports ./random.rn.js as specified in package.json for RN environment.

Here I mean my PR. What happens without my changes in debug mode is: `index.browser.js` gets executed and an exception is thrown (not taking in account the presence of `expo-random` in the project).

I also corrected the test, as `index.browser.js` is now not executed in the react-native context.

ai:
Sorry, I have a lack of RN experience and need more details to change RN behavior (and my answers is not quick, because of vacation).

> index.browser.js gets executed and an exception is thrown (not taking in account the presence of expo-random in the project).

Did I understand correctly, that you load `nanoid` in your RN project, but in debug mode (which works in Chrome) you do not need to use `expo-random`?

But wy do you import `nanoid` and not `nanoid/async`?

skaldo:
No worries! We do have a patch in our repo so this is not that urgent.

Actually your last sentence brought me to the point - I checked the stack trace again and saw `browserLoginFlow.js`. Searching through dependencies I found an internal dependency that is importing `nanoid` and not `nanoid/async`. This explains the mystery how at all `index.browser.js` gets executed.

Why this is not working in debug mode? Checking the code of our dependency reveals following:
```
var crypto = window.crypto || window.msCrypto;
var nanoid = crypto ? require('nanoid') : require('nanoid/non-secure'); 
```
If crypto is available (RN in debug mode), `nanoid` is required -> `index.browser.js` is executed and exception is thrown. In the other case, `nanoid/non-secure` is used.

Seeing all of this, I do not think that you need my PR as this should not happen to a regular user of this library. I would maybe just make the error more explicit. Something like this:
`
You should not import nanoid directly in React Native. If you don't need unpredictable IDs, you should use either nanoid/non-secure or install expo-random and use nanoid/async.
`

If you want, I can update my PR. In the other case, you can just close it.

ai:
It is always a good feeling when you find a reason.

Do you have any idea how we can improve error message to help future developers to find a problem?

swernerx:
Interesting that this came up again. See my last PR which was closed because nobody in our team had time to check: https://github.com/ai/nanoid/pull/136

PR #164, opened 2020-01-07
docs: add comments for all places which are not obvious

Beraliv:
Add comments for all places which are not obvious 

<a href="https://twitter.com/andrey_sitnik/status/1214640079999655936">Link to Russian comment in Twitter</a>
<details>
  <summary>Screenshot of Russian comments in Twitter</summary>
  Screenshot 2020-01-08 at 00 38 47
</details>

### DIFF
--- a/async/format.browser.js
+++ b/async/format.browser.js
@@ -1,4 +1,5 @@
 module.exports = function (random, alphabet, size) {
+  // TODO
   var mask = (2 << 31 - Math.clz32((alphabet.length - 1) | 1)) - 1
   // -~i => i + 1 if n is integer number
   // -~f => Math.ceil(f) if n is float number
--- a/async/format.browser.js
+++ b/async/format.browser.js
@@ -9,6 +9,7 @@ module.exports = function (random, alphabet, size) {
     return random(step).then(function (bytes) {
       var i = step
       while (i--) {
+        // '' is to refuse numbers bigger than the alphabet
         id += alphabet[bytes[i] & mask] || ''
         if (id.length === +size) return id
       }
--- a/async/format.browser.js
+++ b/async/format.browser.js
@@ -5,8 +5,15 @@ module.exports = function (random, alphabet, size) {
   // which is close to `Math.pow(x, 2)`
   // e.g. if we have 7 symbols in the alphabet, bitmask will pass 8 anyway
   var mask = (2 << 31 - Math.clz32((alphabet.length - 1) | 1)) - 1
-  // -~f => Math.ceil(f) if n is float number
-  // -~i => i + 1 if n is integer number
+  // 1. -~f => Math.ceil(f) if n is float number
+  // 2. -~i => i + 1 if n is integer number
+  // 3. It means how much bytes we generate in one round.
+  // Hardware random call generator has a big price
+  // because we need to wait for entropy collection
+  // Hardware random generator needs entropy only for first random seed,
+  // then it generates next bytes by the algorithm.
+  // This is why it is cheaper sometimes to ask for more bytes in the round to reduce calls.
+  // 4. `1.6` is a selected factor of how many extra bytes it's better to ask
   var step = -~(1.6 * mask * size / alphabet.length)
 
   function tick (id) {
--- a/async/format.browser.js
+++ b/async/format.browser.js
@@ -1,8 +1,8 @@
 module.exports = function (random, alphabet, size) {
   // TODO
   var mask = (2 << 31 - Math.clz32((alphabet.length - 1) | 1)) - 1
-  // -~i => i + 1 if n is integer number
   // -~f => Math.ceil(f) if n is float number
+  // -~i => i + 1 if n is integer number
   var step = -~(1.6 * mask * size / alphabet.length)
 
   function tick (id) {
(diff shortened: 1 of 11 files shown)


async/format.browser.js
```diff
@@ -1,5 +1,8 @@
 module.exports = function (random, alphabet, size) {
+  // TODO
   var mask = (2 << 31 - Math.clz32((alphabet.length - 1) | 1)) - 1
```
Beraliv: Should I add anything for masks?
Below as well

non-secure/generate.js
```diff
@@ -18,6 +18,7 @@ module.exports = function (alphabet, size) {
   size = size || 21
   var id = ''
   while (size--) {
+    // TODO
     id += alphabet[Math.random() * alphabet.length | 0]
```
Beraliv: Well, `| 0` isn't obvious for me

async/format.browser.js
```diff
@@ -1,5 +1,8 @@
...
+  // -~i => i + 1 if n is integer number
```
ai: Do we need a note about `integer` here?

async/format.browser.js
ai: 1. We need to refuse numbers bigger than the alphabet. You can add this note to `id += alphabet[bytes[i] & mask] || ''` as well, `|| ''` is to refuse byte if it is bigger than `alphabet.length`
2. To refuse fewer numbers we apply bitmask. For instance, we got to have 8 symbols abt alphabet and random generator gave us `9` (`101`). Without mask hack, we need to refuse this byte. But we apply `11` mask on it and got `001` (`1`).
3. It doesn’t solve all problems, because bitmask can reduce big bytes to closed `x^^2` number (if we have 7 symbols in the alphabet, bitmask will pass `8` anyway).

async/format.browser.js
Beraliv: Just found the comment about it in https://github.com/ai/nanoid/pull/161#issue-359831645
It can be important

async/format.js
```diff
@@ -29,6 +29,7 @@
  * @function
  */
 module.exports = function (random, alphabet, size) {
+  // TODO
   var mask = (2 << Math.log(alphabet.length - 1) / Math.LN2) - 1
   var step = Math.ceil(1.6 * mask * size / alphabet.length)
```
ai: `step` is how much bytes we generate in one round. Hardware random call generator has a big price because we need to wait for entropy collection.

Hardware random generator needs entropy only for first random seed, then it generates next bytes by the algorithm. This is why it is cheaper sometimes to ask for more bytes in the round to reduce calls.

1.6 manually selected constant of how many extra bytes it is better to ask.

async/index.browser.js
```diff
@@ -11,6 +11,7 @@ module.exports = function (size) {
   var id = ''
   var bytes = crypto.getRandomValues(new Uint8Array(size))
   while (size--) {
+    // 63 means last 6 bits
```
ai: Let’s add, that we need to use `00111111` bitmask because we have 64 symbols in the default alphabet.

Since 64 is `x^^2` number, our bitmask hack will work perfectly without any bytes bigger than the alphabet. This is why we do not need `|| ''` and `1.6` hacks here.

non-secure/generate.js
ai: `| 0` is Math.round

index.browser.js
```diff
@@ -27,6 +27,7 @@ module.exports = function (size) {
   var id = ''
   var bytes = crypto.getRandomValues(new Uint8Array(size))
   while (size--) {
```
ai: Let's make a note that `while (size--) {` is a more compact alternative for `for (var i = 0; i < size; i++)`

ai:
Let’s add notes about `random.js` we reuse a single buffer to avoid memory fragmentations

async/index.browser.js
Beraliv: Added changes ✅

index.browser.js
Beraliv: Added ✅

async/format.browser.js
Beraliv: @ai Do you mind if I remain it as is?

ai: Yeap, just move it below. More relevant case (float) first.

Beraliv: Done ✅

non-secure/generate.js
Beraliv: Done  ✅

Beraliv:
> Let’s add notes about `random.js` we reuse a single buffer to avoid memory fragmentations

Done ✅

async/format.browser.js
Beraliv: Done ✅

async/format.js
Beraliv: I didn't change anything because it sounds precise
Done ✅

Beraliv:
@ai have a look again please, all comments are done

ai:
You have an ESLint error: 

```
This line has a length of 92. Maximum allowed is 80  max-len
```

Beraliv:
> You have an ESLint error:
> 
> ```
> This line has a length of 92. Maximum allowed is 80  max-len
> ```

Yeah, already fixed

ai:
Great. I will accept it this evening (by SF time)

Beraliv:
> Great. I will accept it this evening (by SF time)

Thank you for the quick response!

PR #166, opened 2020-01-20
The same behavior in nodejs and browsers.

reklatsmasters:
I found that `nanoid` has different behavior in nodejs and browsers. Details:

## Before

* nodejs, sync
```
nanoid('10') // throw TypeError (test was broken)
			 // works fine if it was called after nanoid(10)
  			 // this is possible due to buffers[10] === buffers['10']
nanoid('foobar') // throw TypeError
```

* nodejs, async
```
nanoid('10') // throw TypeError always
nanoid('foobar') // throw TypeError
```

* browser, sync
```
nanoid('10') // works as nanoid(10)
nanoid('foobar') // return empty string
				 // new Uint8Array('foobar') allocates empty array
				 // while('foobar'--) {} is false
```

* browser, async
works as sync

## After

All versions works like:

```
nanoid('10') // works as nanoid(10)
nanoid('foobar') // works as nanoid()
```

It cost 2 bytes.

---

`nanoid/random` also has differend behavior, but i leave it for next pr / discussion.

### DIFF
--- a/async/index.browser.js
+++ b/async/index.browser.js
@@ -9,7 +9,7 @@ var crypto = self.crypto || self.msCrypto
 var url = 'IUint8Ar21ModulvezGFYPCJ7_p0V4XSymbLBNH6fTqQ35xD9ZREghasOw-cjkWK'
 
 module.exports = function (size) {
-  size = size || 21
+  size = size | 0 || 21
   var id = ''
   var bytes = crypto.getRandomValues(new Uint8Array(size))
   // Compact alternative for `for (var i = 0; i < size; i++)`
--- a/async/index.js
+++ b/async/index.js
@@ -21,7 +21,7 @@ var url = require('../url')
  * @function
  */
 module.exports = function (size) {
-  size = size || 21
+  size = size | 0 || 21
   return random(size).then(function (bytes) {
     var id = ''
     // Compact alternative for `for (var i = 0; i < size; i++)`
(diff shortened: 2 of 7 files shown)


ai:
Here is my philosophy for cases like this:
1. We can warn the developer about the mistake only in the development version if error came from the wrong code
2. We should not hide the mistake from wrong code

I think wrong types are mostly “wrong code” rather than “wrong data” case since it is very unlikely that error will be only in production and not in development.

Because of rule 2, I do not like `| 0` here. If you pass `string`, something wrong happened and you need to fix the code.

What do you think about `if (process.env.NODE_ENV === 'production' && typeof size !== 'number') { throw new TypeError` to `index.browser.js` to show error on wrong type in browser development mode?

reklatsmasters:
* I agree, end user should control your data.
* i was fast low-cost (2 bytes) fix. Nodejs has type check under the hood, i'm ok if you want to do the same check in browsers. The goal of this pr - predictable behavior in every env.

ai:
> i was fast low-cost (2 bytes) fix

We have time to find the best solution.

`if (process.env.NODE_ENV === 'production' && typeof size !== 'number') { throw new TypeError` will cost us 0 bytes and will not hide the wrong types error.

Do you want to update PR?

ai:
I am closing because we made a big refactoring

PR #167, opened 2020-01-20
Non-secure: Generate alphabet using fromCharCode

subzey:
102 -> 89 (-13) bytes for `non-secure/index.js`

----

This can go down to 87 if the `url` variable is reused as `i`:

```
var url =  String.fromCharCode(
  url = 48, url += 1, url += 1, ...
```

There are downsides though:

- eslint rants about a _variable is used before before it was defined_.

- This way the `url` variable changes its type from `number` to `string` in the runtime. At least V8 could JIT the code worse when it encounters a non-monomorphic variable.

PR #168, opened 2020-01-21
Use fromCharCode to generate the alphabet everywhere

subzey:
A #167 follow-up. Apply the same alphabet generation method everywhere else.

-10 bytes for index.js!


ai:
The biggest size reduction!

ai:
Released in 2.1.10

PR #198, opened 2020-03-26
Improve README content

aidarkhanov:
I hope we are hitting this shiny 3.0 version


aidarkhanov:
Is the lower performance nanoid hosted on jsdelivr still relevant with the new nanoid.js file?

PR #202, opened 2020-03-28
do not reference secure js in non-secure js

iamfat:
do not reference secure js in non-secure js, since it will force the bundler to require `crypto`.

### DIFF
--- a/non-secure/index.js
+++ b/non-secure/index.js
@@ -1,4 +1,5 @@
-let { urlAlphabet } = require('..')
+let urlAlphabet =
+  '_-0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
 
 let customAlphabet = (alphabet, size) => {
   return () => {


ai:
What bundler do you use?

Rollup, webpack, and Parcel should follow the `package.json`’s `browder` field and load `index.browser.js` instead. We even [test](https://github.com/ai/nanoid/blob/master/package.json#L84) it with [Size Limit](https://github.com/ai/size-limit).

If this system doesn't work in some bundler we need to fix it.

TrySound:
Using nanoid/non-secure in react-native may give this result as there is no crypto package there.

farwayer:
`import {nanoid} from 'nanoid/non-secure'`

2020-03-28-220200_526x160_scrot

farwayer:
But it is better to find another solution than duplicating `urlAlphabet` as in PR of course.

ai:
React Native fix was released in 3.0.1.

But we still need to find a problem with `packages.json`→`browser`. Does anymore have the same problem of using big Node.js version in client-side JS bundle?

PR #203, opened 2020-03-29
Update index.d.ts

dylanirlbeck:
Since this parameter doesn't appear to be optional for the `customAlphabet` function, we should either remove the `The default size is 21.` from the documentation or update the code to default to a size of `21` if it's not passed in. This PR does the former, but the intent is more to prompt a discussion. Also, please let me know if I'm missing something 😄

Thanks in advance!


ai:
Thanks. Yeap, this fix is correct, I copied/pasted the wrong line.

Your fix was released at 3.0.2.

PR #209, opened 2020-04-12
fix: avoid require bug in node.js 13.0

bluelovers:
https://github.com/nodejs/modules/issues/446

    "Error: Package exports for '/data/data/com.termux/files/home/.config/yarn/global/node_modules/nanoid' do not define a valid '.' target\n"

Screenshot_20200412-121017_Termux

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   "author": "Andrey Sitnik <andrey@sitnik.ru>",
   "license": "MIT",
   "repository": "ai/nanoid",
+  "main": "./index.js",
   "browser": {
     "./index.js": "./index.browser.js",
     "./async/index.js": "./async/index.browser.js"


ai:
@bluelovers [`dual-publish`](https://github.com/ai/dual-publish/) automatically [inserts](https://github.com/ai/dual-publish/blob/master/process-dir.js#L131) `main` to `package.json` during the publish.

How I can reproduce your issue?

bluelovers:
this pr is failed, it can't fix this issue

but if i edit inretface-datastore/src/key.js

replace `nanoid` to `nanoid/index` will not have this error at start up
but i can't replace it at every where

---


install termux on android phone
install git and nodejs , yarn
nodejs will be 13.0.0 <== new version didn't have this issue, but termux only have this version

```bash
yarn global add novel-opds-now
novel-opds-now
```

don't use npm install this
will see this error



---

i also try request termux update nodejs too, but not sure they will see my email xd

ai:
It happens because Node.js 13 does not support conditional exports. Look at `node_modules/nanoid/package.json`.

You can enable it manually by running `node --experimental-conditional-exports`.

TrySound:
@ai But it does support. Just emits warning.

ai:
@TrySound there were some bugs in old Node.js 13.0 versions.

I think we should support only the latest version of each Node.js 10.x, 12.x, and 13.x branches.

Issue #215, opened 2020-05-02
Error: No valid exports main found for '/path/to/nanoid'


cansin:
As initially described at https://github.com/ai/nanoid/issues/206#issuecomment-622967848 , I am getting a `MODULE_NOT_FOUND` error with `node@13.2.0` and `yarn@1.22.4`. 

Here is the repro: https://github.com/cansin/nanoid-module-not-found-repro

And a sample run on my local: 

```bash
cansin@localhost nanoid-repro % node --version
v13.2.0
cansin@localhost nanoid-repro % yarn --version
1.22.4
cansin@localhost nanoid-repro % yarn install
yarn install v1.22.4
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
[4/5] 🔗  Linking dependencies...
[5/5] 🔨  Building fresh packages...

✨  Done in 0.09s.
cansin@localhost nanoid-repro % node index.js 
internal/modules/cjs/loader.js:614
  throw e;
  ^

Error: No valid exports main found for '/Users/cansin/code/nanoid-reprod/node_modules/nanoid'
    at resolveExportsTarget (internal/modules/cjs/loader.js:611:9)
    at applyExports (internal/modules/cjs/loader.js:492:14)
    at resolveExports (internal/modules/cjs/loader.js:541:12)
    at Function.Module._findPath (internal/modules/cjs/loader.js:643:22)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:941:27)
    at Function.Module._load (internal/modules/cjs/loader.js:847:27)
    at Module.require (internal/modules/cjs/loader.js:1016:19)
    at require (internal/modules/cjs/helpers.js:69:18)
    at Object.<anonymous> (/Users/cansin/code/nanoid-reprod/index.js:1:20)
    at Module._compile (internal/modules/cjs/loader.js:1121:30) {
  code: 'MODULE_NOT_FOUND'
}
```
Thanks.

TrySound:
Now, try to upgrade node.

cansin:
Btw, either upgrading to the latest Node, or even downgrading to LTS `12.16.3` fixes this issue. So perhaps it is not related to the library. But still, you might want to update your engines.node to a valid list.

cansin:
@TrySound see above ^ 😊

ai:
Thanks for the repo. I will take a look today or tomorrow.

[...]

PR #216, opened 2020-05-04
Restrict failing node@13.x versions at engines.node

cansin:
~~Since @ai mentioned we only support `^10`, `^12`, and `>=13`, I didn't include `^11`. But if there is no reason not to support `^11` (as far as I can tell it actually works), we can update the semver as `">=10 <13 || >=13.7"`.~~

Nvm, I didn't realize Node disowned `11` 😂 

Alright this fixes https://github.com/ai/nanoid/issues/215

### DIFF
--- a/package.json
+++ b/package.json
@@ -186,5 +186,8 @@
   "sharec": {
     "config": "@logux/sharec-config",
     "version": "0.6.2"
+  },
+  "engines": {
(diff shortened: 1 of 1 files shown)


ai:
Thanks. I will release it in next hour.

cansin:
Take your time. Thanks again.

ai:
Released in 3.1.4

PR #218, opened 2020-05-08
Typo in README

dimitrinicolas:
Changed `key` for `id` on the React usage part about HTML ID that cannot start with a number.


ai:
Thanks

PR #219, opened 2020-05-09
Add Typegoose usage

dandv:
Thanks for the crazy cool module!

Would it be Ok to submit a PR for [retrying saves in case of collisions](https://github.com/typegoose/typegoose/issues/159#issuecomment-626079521), or would that be out of scope?


ai:
We need some policy about what to include. But let’s add it for now.

ai:
Just fix CI issues.

PR #220, opened 2020-05-14
Fix customAlphabet call causing infinite loop if called without size param

DKhalil:
…param

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,7 +54,7 @@ let customRandom = (alphabet, size, getRandom) => {
   }
 }
 
-let customAlphabet = (alphabet, size) => customRandom(alphabet, size, random)
+let customAlphabet = (alphabet, size = 21) => customRandom(alphabet, size, random)
 
 let nanoid = (size = 21) => {
   let bytes = random(size)


ai:
We didn’t add default `size` for a reason, because the recommended size depends on the alphabet.

Also, I do not want to add arguments check, because there is TypeScript for type checks (and we provide great TS support).

Was it really painful bug for you, or it was just “find and fix”?

DKhalil:
It was rather painful, since it just gets stuck in an infinite loop and doesn't break in any predictable way - which in a larger application can be pretty difficult to find. So if you don't want to have a default size, maybe just exit the function with an error if the size param isn't provided instead?

ai:
> So if you don't want to have a default size, maybe just exit the function with an error if the size param isn't provided instead?

It will increase the size of the bundle.

We had our in 2.x, but then I decided to move all arguments checks to the TypeScript type definitions.

DKhalil:
I mean, this kind of check is probably going to weigh a few bytes in the final bundle, plus what about people who don't use typescript?

ai:
I do not use TS for most of my projects too. But I think, that if you use non-types language, that you must check all signatures manually.

If you have signatures auto-check you should use proper types language, rather than adding hacks to libraries.

DKhalil:
What signatures?

I'm just saying, an improper or missing function argument shouldn't cause an infinite loop and total failure without notice or warning. Does that not make sense to you? Am I missing something here? In my opinion bundle size shouldn't just completely trump over proper API/UX design. And in this case, calling customAlphabet('abcdef1235') like that, would simply crash the application entirely.

Now, if you disagree with that and allowing your users to run into a brick wall is a consequence you are willing to allow, then let's be done with it and close this and maybe add a warning to the documentation about custom alphabets then.

ai:
> I'm just saying, an improper or missing function argument shouldn't cause an infinite loop and total failure without notice or warning.

If you will use the library wrong you will crash the app in any case. There are a lot of ways to use it wrong.

We can improve the DX by checking user input and showing a more relevant error message. But I decided to stop doing it, since we have TypeScript. If you want the system to check how do you use API, use TypeScript.

PR #229, opened 2020-07-24
async/customAlphabet - ensure browser version returns Promise

### LINKED ISSUE #228: customAlphabet in nanoid/async/index.browser is not actually async
Visual comparison of  `customAlphabet()` between `/index.browser.js` and `/async/index.browser.js` suggests that the async version is really synchronous. This should either be changed to at least return a Promise, or documented.

LoneRifle:
To ensure consistency in the functions exported by async/browser.index
and to not catch out users expecting a Thenable returned by functions
claimed to be async, ensure that the final value returned by 
`customAlphabet()` is a Promise resolving to the value, not the value
itself

Fixes #228


ai:
Thanks. Released in 3.1.11.

PR #232, opened 2020-09-11
Update README.md

Doogiemuc:
Added hint for custom alphabet usecase


ai:
I think we have a special solution for this type of IDs. Maybe we should show it in example about this paragraph?

Can you check the Tools section and add an example or link?

Doogiemuc:
Ok I see. Perfect. Thanks!

Issue #237, opened 2020-10-01
ES module browser build expects "process" global


guybedford:
When running this package on jspm, it will give "process is undefined" because of the process check in the index.browser.js file.

The assumption of the browser environment shouldn't be based on the proces global existing. Adding a `typeof process !== 'undefined'` check would fix this issue.

ai:
@guybedford can you send PR with `typeof process !== 'undefined'` to save your time in the history?

TrySound:
process.env.NODE_ENV will never be reached because of `typeof process` for example in rollup.
This is widely used convention and it's not specific to node packages. "browser" field does not mean process.env.NODE_ENV cannot be used.

guybedford:
@TrySound to be super clear here, while `index.browser.js` in this repo is CommonJS, the published version is published as an ES module - https://unpkg.com/nanoid@3.1.12/index.browser.js. It is the combination of assuming the process global, setting this as the entry for the browser platform, and then also shipping it as an ES module that is specifically the issue here. Any of the lesser combinations wouldn't be as bad!

ai:
As a temporary solution you can use `import { nanoid } from "nanoid/nanoid.js"`

guybedford:
@ai the problelm is a dependency of a dependency importing `nanoid` is broken when respecting the `"browser"` exports field breaking all packages on jspm which depend on nanoid. It would be preferable not to add the `"browser"` field if it isn't going to be a valid ES module for the browser.

ai:
Let’s create an issue in jspm to add `process.env.NODE_ENV` support

guybedford:
@ai the problem is that `process` is not a browser global - therefore an ES module that expects it to exist will break in the browser. jspm follows browser semantics so will not fix this.

guybedford:
Only when using CommonJS will jspm support the `process` global by default.

ai:
@guybedford do you see any other option on how to keep good DX and support browser API?

I am suggesting changing jspm because I do not see other options.

guybedford:
@ai I'm not sure why the original form of this PR to add `typeof process` wasn't a suitable option to be honest.

ai:
Because webpack doesn't add `process` to the system. It looks for the exact `process.env.NODE_ENV` call and replace only this code to `"production"` or `"development"` string:

```diff
if (typeof process !== "undefined" && process.env.NODE_ENV === "development") {
```

Will be compiled to:

```diff
[...]

PR #238, opened 2020-10-01
Add process definition check

guybedford:
Fix as per https://github.com/ai/nanoid/issues/237, pending further testing.

### DIFF
--- a/index.browser.js
+++ b/index.browser.js
@@ -3,7 +3,7 @@
 
 let { urlAlphabet } = require('./url-alphabet')
 
-if (process.env.NODE_ENV !== 'production') {
+if (typeof process !== 'undefined' && process.env.NODE_ENV !== 'production') {
   // All bundlers will remove this block in the production bundle.
   if (
     typeof navigator !== 'undefined' &&
--- a/index.browser.js
+++ b/index.browser.js
@@ -1,4 +1,4 @@
-if ('process' in globalThis && process.env.NODE_ENV !== 'production') {
+if (process.env.NODE_ENV !== 'production') {
   // All bundlers will remove this block in the production bundle.
   if (
     typeof navigator !== 'undefined' &&
--- a/index.browser.js
+++ b/index.browser.js
@@ -3,7 +3,7 @@
 
 let { urlAlphabet } = require('./url-alphabet')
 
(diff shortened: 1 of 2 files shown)


ai:
Adding `typeof process` force webpack to add a huge `process` polyfill and bloat the library size 7x times.

Try `"process" in this` or some similar hack.

TrySound:
Please see my comment
https://github.com/ai/nanoid/issues/237#issuecomment-702338527

guybedford:
Hmm actually I don't think there's a way to accomplish this if the constraint is Webpack support!

@ai note that `this` is not the global in native ES modules, it is instead `.undefined`

@TrySound that's fine but assumes that all ES modules should run through a build system to execute in an environment that is not Node.js. We shouldn't have to enforce this restriction on the JS ecosystem, and that's an important thing to avoid lock-in to tooling silos over native workflows which should define the base-level semantics. The tail is wagging the dog currently!

Posted https://github.com/webpack/webpack/issues/11569.

ai:
@guybedford maybe we need to create an issue in jspm? `process.env.NODE_ENV` is a popular way.

With `typeof process` we do not solve the problem. jspm build still contains unnecessary code for production (we just disable it, but still send it to client).

guybedford:
@ai the modern way to do this branching is with the `"exports"` field:

```json
{
  "exports": {
    "production": "./production-index.js",
    "development": "./development-index.js"
  }
}
```

and jspm supports this.

By assuming a build tool will inline `process.env.NODE_ENV` for a native ES module, that means that the "browser" export defined here is not really a browser file at all - but rather a module that is expected to run through a bundler. It is that expectation which continues to create a reliance on overly complex build tools for beginners to use the web platform, when modules should be able to execute in browser directly with just a local server and an import map, which is a key use case jspm is designed to enable and not restrict. Tools should be designed to set users free of them.

ai:
@guybedford we can add `exports.production: 'index.production.js'` and put a copy of `index.browser.js` without a development code

guybedford:
@ai the problem with that is that even then, jspm will use the "development" branch on jspm.dev which will still break!

I'd suggest shipping CommonJS for this file if assuming the process global, and then later on switching to "development" / "production" exports once those have wider support in tooling.

PR #242, opened 2020-10-25
Example has 22 not 21 chars

phoenixbox:
Seems to be a typo, is the limit 21 chars?


ai:
Yeap. Could be of example. Thanks.

PR #252, opened 2020-12-14
fix(README.md): Deleted description mentioning the incorrect way to use nanoid with React elements' keys.

FokinAleksandr:


ai:
We can't remove the example because we need to explain what is the right way to do it

README.md
```diff
@@ -191,22 +191,36 @@ import { nanoid } from 'nanoid'
...
+        </li>
+      ))}
+    </ul>
+  )
+}
 ```
-
-This is the good example (`id` will be generated only once):
-
+You should rather try to reach for stable id inside your list item
```
ai, line 209: We had newline between text and code examples

README.md
```diff
@@ -191,22 +191,36 @@ import { nanoid } from 'nanoid'
...
-  const [id] = React.useState(nanoid)
-  return <Item key={id} />
-}
+const todoItems = todos.map((todo) =>
+  <li key={todo.id}>
+    {todo.text}
+  </li>
+);
+```
+In case you don't have stable ids you'd rather use index as key instead of nanoid();
```
ai, line 217: Let’s keep the original good typography:

```suggestion
In case you don’t have stable ids you'd rather use index as `key` instead of `nanoid()`:
```

README.md
```diff
@@ -191,22 +191,36 @@ import { nanoid } from 'nanoid'
...
+```jsx
+const todoItems = todos.map((todo, index) =>
+  <li key={index}> /* Still not recommended but prefered over nanoid(). Only do this if items have no stable IDs */
+    {todo.text}
+  </li>
+);
```
ai, line 223: We do not use `;` in other examples

ai:
Thanks

PR #255, opened 2020-12-22
add tool

junkboy0315:
Hi, I created following online tool that easily generates a large number of Nano ID using presets.
[https://nanoid.yuuniworks.com/](https://nanoid.yuuniworks.com/)

I would be grateful if you could add this to the README.
(I'm worrying if I'm allowed to place ads though)


ai:
When I open websit, I got “We've noticed that you are using an ad blocker” message. But I don’t have ad blockers. I just use Firefox.

I can’t add a link to service which force users to disable privacy tools.

PR #263, opened 2021-02-10
Improve Rollup usage documentation

hudochenkov:
I have an React component library, which is bundled with Rollup. Then this library is used with app bundled with webpack. After importing `nanoid` in this library and bundling with Rollup. Webpack then adds 590.15kb of Node.js polyfill to the bundle.

The reason is Rollup bundles `index.js`, which has `import crypto from 'crypto'`. When I configure Rollup `node-resolve` plugin to search for `browser` field, then `index.browser.js` imported into bundle.

I thought update of documentation could help other developers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,17 +273,22 @@ For Expo framework see the next section.
 
 ### Rollup
 
-For Rollup you will need [`@rollup/plugin-replace`] to replace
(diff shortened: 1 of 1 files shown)


ai:
Yeap. Useful improvments. Thanks.

PR #284, opened 2021-06-12
Stabilize tests in async/index.js

### LINKED ISSUE #283: Stabilize tests
`jest --coverage` depends on random and sometimes ignores 50 line in `async/index.js`

Example: https://github.com/ai/nanoid/runs/2749571294?check_suite_focus=true

immitsu:
Fix #283

### DIFF
--- a/async/index.js
+++ b/async/index.js
@@ -38,19 +38,19 @@ let customAlphabet = (alphabet, size) => {
   // according to benchmarks).
   let step = Math.ceil((1.6 * mask * size) / alphabet.length)
 
-  let tick = id =>
-    random(step).then(bytes => {
-      // A compact alternative for `for (var i = 0; i < step; i++)`.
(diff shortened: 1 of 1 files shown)


ai:
Can I ask you to run `./test/benchmark` before and after these changes?

immitsu:
Sure.
I've run both versions five times, and the old `async customAlphabet`, on average, is a little bit faster - 69.1 against 68.5

ai:
New change will not work. We do not know how many random bytes we will need (because we are ignoring done sets). This is why we have `while (true)`.

ai:
Performance drop is not a good thing. Especially because we can fix this issue by fixing the test, not the code

ai:
The problem was fixed with `crypto` mock https://github.com/ai/nanoid/pull/285

PR #286, opened 2021-06-23
Fix NodeJs example in README

theproductiveprogrammer:
The README shows

```js
const { nanoid } = require('nanoid')
```

which causes the call `nanoid()` to crash in NodeJS with the error: `TypeError: nanoid is not a function`

Fixed by removing the braces

```js
const nanoid = require('nanoid')
```


ai:
Nope. The correct form in Nano ID 3 is `{ nanoid }`.

Maybe you are using old Nano ID 2?

theproductiveprogrammer:
Ah you're right - it was nano id 2.1.11

PR #303, opened 2021-09-25
avoid 'subarray' call when possible in sync implementation

eharshosh:
As title says - avoiding call to `pool.subarray` improves performance significantly for the `nanoid` method in the synchronous implementation.
running benchmark.js on my machine before yield the following:
```
nanoid                    1,371,251 ops/sec
customAlphabet            1,017,790 ops/sec
```
and after:
```
nanoid                    1,954,770 ops/sec
customAlphabet            1,143,370 ops/sec
```
* added `customAlphabet` here since this changes adds a method call and should've slowed it down a little, but as the benchmark shows - it is negleggible and the benchmark itself doesn't seem to be accurate enough to catch this change

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,11 @@ const POOL_SIZE_MULTIPLIER = 32
 let pool, poolOffset
 
 let random = bytes => {
+  _fillPool(bytes)
+  return pool.subarray(poolOffset - bytes, poolOffset)
+}
+
+let _fillPool = bytes => {
   if (!pool || pool.length < bytes) {
     pool = Buffer.allocUnsafe(bytes * POOL_SIZE_MULTIPLIER)
     crypto.randomFillSync(pool)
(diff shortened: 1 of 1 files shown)


ai:
Amazing! Do you have Twitter account to mention you in release tweet?

index.js
```diff
@@ -63,16 +65,15 @@ let customRandom = (alphabet, size, getRandom) => {
 let customAlphabet = (alphabet, size) => customRandom(alphabet, size, random)
 
 let nanoid = (size = 21) => {
-  let bytes = random(size)
+  _fillPool(size)
   let id = ''
-  // A compact alternative for `for (let i = 0; i < size; i++)`.
-  while (size--) {
+  for (let i = poolOffset - size; i < poolOffset; i++) {
```
ai, line 70: We do not update `poolOffset` after using the random value.

Will it lead to reusing values from the pull between `nanoid` calls?

eitanhs:
Thanks!
Nit very active. it would be nice if you mention my GitHub account though :)

PR #304, opened 2021-09-27
Create dependabot config file.

medz:


ai:
Why do we need it?

medz:
@ai It can help us update our dependencies in a timely manner, and make our energy more focused on writing code, instead of having to update the dependencies for a long time and remember to update the dependencies. In addition, the packages we rely on may have security issues, and timely updating of dependencies can be easy to solve.

ai:
Sorry, I prefer to update dependencies manually.

PR #316, opened 2021-11-04
Remove HTML `id` advice

tigt:
HTML5 lifted the restrictions on `id` values, which are now essentially any character sequence: https://mathiasbynens.be/notes/html5-id-class

This was specced about 11 years ago and made with HTML ≤4.01 backwards compatibility in mind. It should be safe to no longer worry about this.


ai:
Cool! Great to know.

PR #319, opened 2021-11-28
docs: remove notes about HTML id usage

andrewvasilchuk:
Completely removed notes about HTML id usage. "HTML ID" anchor was broken in EN docs since this section was deleted in https://github.com/ai/nanoid/pull/316.


ai:
Thanks

PR #326, opened 2021-12-30
Add documentation in Bahasa Indonesia

lauslim12:
Hi!

Thanks for making this library. It's super simple, fast, and very useful for my use-cases!

If you don't mind, I have translated and added `README.id-ID.md` for internationalization purposes (I'm from Indonesia). The documentation now has four languages: English, Russian, Chinese, and Bahasa Indonesia 🎉

By the way, I have a question about a single statement in `async` part of the documentation:

> Unfortunately, you will lose Web Crypto API advantages in a browser if you use the asynchronous API. So, currently, in the browser, you are limited with either security or asynchronous behavior.

The 'security' part here is a bit confusing for me. So, in order to remove any ambiguity, I think it's better if we rewrite this to be like:

> Unfortunately, you will lose Web Crypto API advantages in a browser if you use the asynchronous API. So, currently, in the browser, you are limited with either security (using `nanoid`), asynchronous behavior (using `nanoid/async`), or non-secure behavior (using `nanoid/non-secure`) that will be explained in the next part of the documentation.

I have not yet edited it, but what do you think about this rewrite?

Hopefully, my PR is useful. Thanks again!

### DIFF
--- a/README.id-ID.md
+++ b/README.id-ID.md
@@ -1,0 +1,410 @@
+# Nano ID
+
+<img src="https://ai.github.io/nanoid/logo.svg" align="right"
+     alt="Logo Nano ID oleh Anton Lovchikov" width="180" height="94">
+
+[English](./README.md) | [Русский](./README.ru.md) | [简体中文](./README.zh-CN.md) | **Bahasa Indonesia**
+
+Sebuah generator ID yang unik dalam bentuk string yang ringan, aman, serta _URL-friendly_ untuk JavaScript.
+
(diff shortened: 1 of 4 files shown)


ai:
I like that language code `id` is relavant to the lib :D

ai:
> I have not yet edited it, but what do you think about this rewrite?

I like the suggestion, please add commit to this PR

lauslim12:
> I like that language code id is relavant to the lib :D

:D

> I like the suggestion, please add commit to this PR

Added 🥳

ai:
Thanks!

PR #328, opened 2022-01-11
fix collision by proxy number

artalar:
You can test live reproduce here https://stackblitz.com/edit/nanoid-collision?file=index.js (type `node index.js` in the console)


ai:
Thanks. Fixed was released in 3.1.31.

PR #330, opened 2022-01-11
Replace Jest to uvu

vitalybaev:
Hi!

To be honest, this PR was created occasionally. I haven't seen issues to replace Jest with `uvu` at `nanoid`, but tried to solve this challenge, and it seems that `uvu` fits here well. As far as I can see, all points are solved: same context, coverage, mocking. But for now this PR is in draft.


.nycrc
```diff
@@ -0,0 +1,10 @@
+{
```
TrySound, line 1: c8 uses .nycrc?

vitalybaev: Yep
> A configuration file can be specified by passing its path on the command line with --config or -c. If no config option is provided, c8 searches for files named .c8rc, .c8rc.json, .nycrc, or .nycrc.json, starting from cwd and walking up the filesystem tree.

But you are right, we should use `.c8rc` in this case.

I'll change it tomorrow

vitalybaev: Moved `c8` config to `c8` section in `package.json` in favor of consistency

.github/workflows/test.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile --ignore-scripts
       - name: Run unit tests
-        run: ./node_modules/.bin/jest
+        run: pnpm test:short
```
ai, line 51: Can you rename it to `pnpm unit` for consistency with my other open source projects?

bin/nanoid.test.js
```diff
@@ -1,11 +1,15 @@
+let { test } = require('uvu')
+let assert = require('uvu/assert')
```
ai, line 2: In my other open source projects I use `let { equal, is } = require()` style. Can you convert it too?

package.json
```diff
@@ -9,7 +9,8 @@
     "url"
   ],
   "scripts": {
-    "test": "jest --coverage && eslint . && check-dts && size-limit",
+    "test:short": "c8 --check-coverage uvu . .test.js$",
```
ai, line 12: Let’s move `c8` to `test` to make `test:short` (or `unit` for better consistency with other projects) faster.

package.json
```diff
@@ -164,16 +157,18 @@
...
-      "global": {
-        "statements": 100
-      }
-    }
-  },
   "clean-publish": {
     "cleanDocs": true,
     "cleanComments": true
+  },
+  "c8": {
```
ai, line 164: Here is my universal config https://github.com/ai/boilerplates/blob/main/esm-lib/package.json#L40-L46

ai:
😍 You make my day happier with this PR. Fewer dependencies and better performance always make a day better.

vitalybaev:
Glad to hear that!

I've pushed fixes for all the review notes

PR #332, opened 2022-01-12
add gitpod config

artalar:


ai:
We had it before, but it was very rare used.

I do not like too many configs for rare cases.

PR #334, opened 2022-01-15
Allows passing size or custom alphabet via cli as args

### LINKED ISSUE #275: CLI Improvements
Ahoy 👋  I was curious if the project would be open to a contribution of some CLI improvements. I know that `nanoid-cli` exists, but it's dang near three years out of date and while I have reached out to the author about updating it, the code behind that package is small enough that I wondered if there's openness now to having that integrated as part of `nanoid` proper.

vitalybaev:
Closes #275 

I've implemented minimal args parsing logic to fit our needs and be as little as possible


bin/parseArgs.js
```diff
@@ -0,0 +1,44 @@
+let cleanArgName = (arg) => arg.startsWith('--') ? arg.slice(2) : arg.slice(1)
```
ai, line 1: Let’s move it to `bin/nanoid.cjs` since we do not use it anywhere

ai:
We need to improve test coverage. You can check test coverage locally by running `pnpm test`.

ai:
We also need to fix Node.js 12 support for binary (we can’t use `??`)

bin/parseArgs.js
vitalybaev: Didn't understand you. `cleanArgName` is being used in `bin/parseArgs.js:16` and `bin/parseArgs.js:23`, so in the same file.
Or we can simply omit such sanitizing at all and just use args keys like `parsedArgs['--size']` (and save several bytes). What do you think?

ai: I suggested moving all `bin/parseArgs.js` content directly to `bin/nanoid.cjs` (the comment is not for the function,. but for the whole file).

vitalybaev: Ah. But I test `parseArgs` function separately. In this case, we have to export this function from `bin/nanoid.cjs`. I don't know exactly whether it is OK. Or is it better to test  only real cli exec?

ai: Hm. Yes, in this case we can keep a separated file.

Let’s rename it to `bin/utils.js`

vitalybaev:
Done

I've faced error ` ERROR  dual-publish error: Rename bin/utils.js to bin/utils/index.js`, so I moved file according to this message.

Now it still works with Node 12 and test coverage is 100%

ai:
`bin/utils` will create unnecessary files because of `dual-publish`. I cleaned up arguments parsing a little https://github.com/ai/nanoid/commit/27eaa90cd207a7782bbcf17343092ae87dd62164

And released in 3.2.

PR #346, opened 2022-02-06
Switch Haskell NanoID implementation to the Community maintained one

MichelBoucey:
I have created and I maintain the [NanoID](https://github.com/MichelBoucey/NanoID), library and cli tool, which can be found on Haskell Community website, [Hackage](https://hackage.haskell.org/package/NanoID). That's why I suggest, with this PR, to replace the current Haskell implementation, which has a pure educational purposes, according to the author himself, by my implementation, built idiomatically the Haskell way.


ai:
@4e6 is it OK? We can mention both if there is a system difference.

4e6:
I'm totally fine with the replacement :+1:

PR #355, opened 2022-04-18
reduce bundlesize by moving from while to reduce

artalar:


ai:
😍😍😍

artalar:
Just forcepushed another commit, thats all now :)

ai:
Thanks. Released in 3.3.3.

levrik:
I'm wondering how the move from `while` to `reduce` affects performance. Method calls tend to be more expensive.

ai:
@levrik it is micro-optimization (especially in the case of browser version, Node.js version uses `while`).

ai:
@levrik I looked in benchmark (`pnpm start` to benchmark browser version) and didn’t find any changed before and after

PR #361, opened 2022-05-03
fix: CLI help examples

Lete114:
I think @vitalybaev made a mistake ([/bin/nanoid.cjs#L19](https://github.com/ai/nanoid/blob/32b9bdaab1fbc28576b17de8516164ce0360f292/bin/nanoid.cjs#L19))


vitalybaev:
Yes, definitely, nice catch 👍

ai:
Thanks. Released in 3.3.4.

PR #371, opened 2022-07-02
docs: new link to Go implementation

jaevor:
I created a newer and improved Go implementation of nanoid ([link](https://github.com/jaevor/go-nanoid)) and I wanted to update the links in the reference repo 🙂.


ai:
Thanks. What is the different between version?

Is it OK for @matoous?

jaevor:
> Thanks. What is the different between version?

Mine is faster. I implemented the random crypto byte prefetching for more efficiency and I use `byte` for the default alphabet instead of `int32` since `A-Za-z0-9_-` is ASCII - and lots more efficient memory.

matoous:
Sure, go ahead if it's faster. 👍

PR #380, opened 2022-08-16
hardcode 'other programming languages' TOC link to fix it on npmjs.com

jgravois:
the existing link is a deadend on https://www.npmjs.com/package/nanoid

if this approach is sound, i'm happy to tack on another commit to fix the translated README files too.


ai:
> translated README files too

Do we need it? We can’t open them from on npm.

PR #384, opened 2022-09-11
Remove and ignore lock files

JairusSW:
Just a quick PR to remove and ignore lock files. 
Typically, lock files are not included in a repository
:)


ai:
I always have lock files in my open source repositories.

Don't worry npm package should not have it. We use `clean-publish` to clean development files and configs from `package.json` during the release.

JairusSW:
Okay, np

PR #385, opened 2022-09-13
docs: use「hack」word

YeeJone:
There is no word「骇客」in Chinese..

### DIFF
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -136,7 +136,7 @@ rndm                       2,674,806 ops/sec
 npm install --save nanoid
 ```
 
-对于快速的骇客用法，你可以从 CDN 加载 Nano ID。但是，它不建议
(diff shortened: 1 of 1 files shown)


ai:
Can we rephrase to avoid English word?

cc @wenqiang @iDestin

iDestin:
> Can we rephrase to avoid English word?
> 
> cc @wenqiang @iDestin

Of course, when I translated at that time, I thought hack was a word that everyone knew, and everyone could know the meaning of it when I used it directly, so I chose not to translate it. It may be a better choice to translate it into Chinese

ai:
@iDestin can you send PR? I'm afraid to do it incorrectly because my eyes is not trained to operate with characters.

ai:
Close for https://github.com/ai/nanoid/pull/386

PR #390, opened 2022-09-22
docs: update Go library to atoous/go-nanoid

hbagdi:
https://github.com/jaevor/go-nanoid license was changed to GPL. Recommend an alternate implementation.


ai:
I don’t have any problem with GPL.

Is there any other difference?

KostiantynPrykhodko (approved):

ItalyPaleAle:
We are using https://github.com/matoous/go-nanoid in Dapr too, as any library with a GPL license is a non-starter sadly (it would force us to release the entire Dapr under GPL).

(Also to say, one could fork jaevor's library from before the license change and continue to have that released with the old license, since license changes are not retroactive)

ItalyPaleAle (approved):

PR #415, opened 2023-03-09
Replace deprecated `expo-random` package

### LINKED ISSUE #414: `expo-random` has been deprecated
Hi there! I just wanted to give you a quick heads up about [`expo-random`](https://docs.expo.dev/versions/latest/sdk/random/) being deprecated. It'll be removed in the next Expo SDK release.

I *think* you'll want to update [index.native](https://github.com/ai/nanoid/blob/main/async/index.native.js) so that the `import` reads:

```
import * as Crypto from 'expo-crypto'
```

And a couple lines below that

```
export let random = Crypto.getRandomBytesAsync
```

silvestreh:
Removes usage of the now deprecated `expo-random`. Closes #414

### DIFF
--- a/async/index.native.js
+++ b/async/index.native.js
@@ -1,8 +1,8 @@
-import { getRandomBytesAsync } from 'expo-random'
+import * as Crypto from 'expo-crypto'
 
 import { urlAlphabet } from '../url-alphabet/index.js'
 
-export let random = getRandomBytesAsync
+export let random = Crypto.getRandomBytesAsync
 
 export let customAlphabet = (alphabet, defaultSize = 21) => {
   // First, a bitmask is necessary to generate the ID. The bitmask makes bytes
--- a/async/index.native.js
+++ b/async/index.native.js
@@ -1,8 +1,8 @@
-import * as Crypto from 'expo-crypto'
+import { getRandomBytesAsync } from 'expo-crypto'
 
 import { urlAlphabet } from '../url-alphabet/index.js'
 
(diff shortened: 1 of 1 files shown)


ai:
Should we release it as Nano ID 5.0?

silvestreh:
🤔 technically speaking, `expo-random` has been merged into `expo-crypto` bringing the functionality over to the newer package. There are no breaking API changes, but then again, people using older Expo SDKs will probably be greeted with an error during build time about the `expo-crypto` not being found.

I believe tagging a major release, even with no new functionality, would make sense.

async/index.native.js
```diff
@@ -1,8 +1,8 @@
-import { getRandomBytesAsync } from 'expo-random'
+import * as Crypto from 'expo-crypto'
```
ai, line 1: Can we avoid importing everything?

```suggestion
import { getRandomBytesAsync } from 'expo-crypto'
```

silvestreh: Just updated the code :)

ai:
I will wait for Node.js 14 to release to make more major changes together.

ai:
@silvestreh BTW, we are using `expo-crypto` only for async version.

Maybe we should use just [browser sync version](https://github.com/ai/nanoid/blob/main/async/index.browser.js) instead with [polyfill](https://github.com/LinusU/react-native-get-random-values) (and wrap it to `Promise`)? Too rare use case for special version.

ai:
Done 0276499

PR #466, opened 2024-01-31
Much faster implementation

samuelgozi:
I know this is a lazy PR since I didn't check other files to make sure I'm making the correct changes, so I'm sorry about that.
This implementation is much faster.
The benchmark checks the time to generate one million nanoids.
```
$ node ./main.js
Old 1365ms
New 948ms

$ bun ./main.js
Old 462ms
New 138ms
```

I hope you find it useful 🙏

### DIFF
--- a/index.browser.js
+++ b/index.browser.js
@@ -48,14 +48,12 @@ export let customAlphabet = (alphabet, size = 21) =>
   customRandom(alphabet, size, random)
 
 export let nanoid = (len = 21) => {
-	let id = "";
-  // This will be hoisted by the JIT engine.
-  // I keep it here to make the code compact and easy to copy.
-	const chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
-	const rand = crypto.getRandomValues(new Uint8Array(len));
+  let id = "";
+  let chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
+  let rand = crypto.getRandomValues(new Uint8Array(len));
   // Using the bitwise AND operator to "cap" the value of
   // the random byte from 255 to 63, in that way we can make sure
   // that the value will be a valid index for the "chars" string.
-	for (let i = 0; i < len; i++) id += chars[rand[i] & 63];
-	return id;
+  for (let i = 0; i < len; i++) id += chars[rand[i] & 63];
+  return id;
 }
--- a/index.browser.js
+++ b/index.browser.js
@@ -47,25 +47,15 @@ export let customRandom = (alphabet, defaultSize, getRandom) => {
 export let customAlphabet = (alphabet, size = 21) =>
   customRandom(alphabet, size, random)
 
-export let nanoid = (size = 21) =>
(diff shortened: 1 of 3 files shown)


ai:
1. It is the file generated from `index.browser.js`. You need to fix it first.
2. But browser's file should be tested in browsers. For Node/Bun we have another file `index.js` which has different optimizations.

samuelgozi:
Tested on latest version:

Chrome
```
Old 1634ms
New 1245ms
```

Safari
```
Old 1208ms
New 760ms
```

Firefox
```
Old 1207
New 742
```

Browser support doesn't change much.
I will update the source file.

samuelgozi:
I now updated the right place.

nanoid.js
```diff
@@ -1 +1 @@
-export let nanoid=(t=21)=>crypto.getRandomValues(new Uint8Array(t)).reduce(((t,e)=>t+=(e&=63)<36?e.toString(36):e<62?(e-26).toString(36).toUpperCase():e>62?"-":"_"),"");
```
ai, line 1: Could you revert changes? I will update it by script.

index.browser.js
```diff
@@ -47,25 +47,15 @@ export let customRandom = (alphabet, defaultSize, getRandom) => {
...
-      id += '_'
-    }
-    return id
-  }, '')
-
+export let nanoid = (len = 21) => {
+	let id = "";
+  // This will be hoisted by the JIT engine.
+  // I keep it here to make the code compact and easy to copy.
+	const chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
```
ai, line 54: Should we move it outside of `nanoid()` function?

index.browser.js
```diff
@@ -47,25 +47,15 @@ export let customRandom = (alphabet, defaultSize, getRandom) => {
...
-    } else if (byte > 62) {
-      id += '-'
-    } else {
-      id += '_'
-    }
-    return id
-  }, '')
-
+export let nanoid = (len = 21) => {
+	let id = "";
```
ai, line 51: 1. Let’s fix code style issues from ESLint.
2. And we will need to update `package→size-limit` to see how much we increase the JS bundle size (for browser’s version we are optimizing both size and performance)

samuelgozi:
Ok i will fix everything you mentioned.
In regards to the use of the string inside the function, I don't think we should hoist it manually since I didn't see any effect on performance, but I do think that this way it looks nicer.

Maybe I should remove the comment instead since it not important

ai:
OK, let's keep string as is. I will do some optimizations later for smaller bundle size.

samuelgozi:
Done. Package size limit has exceeded by 34 B

ai:
> Package size limit has exceeded by 34 B

Try to import alphabet from https://github.com/ai/nanoid/blob/main/url-alphabet/index.js

We have a fun hack there, we put the same letters not in alhabetical order, but in order for better gzip.

And please update `packge→size-limit` for new size (so we will track size changes by commit).

samuelgozi:
`Package size limit has exceeded by 15 B`
TIL, very cool!

Made the changes

ai:
I also reduced 8 bytes by reusing `size` variable 4db3464

samuelgozi:
Using the while loop is actually a little bit faster.
Very nice idea!

PR #474, opened 2024-04-30
Update java library

polvallverdu:
Hey,
A while back I forked the original Java nanoid library (jnanoid), since optimizations and improvements could be made:

- Security Issue: Remote code execution could be ran.
- Random Security: Java's more secure Random wasn't being used.
- Better Unit Testing
- More useful methods

In the readme the recommended Java Library is still pointing to the discontinued one. I've updated it.


ai:
Thanks for keeping that project alive!

PR #485, opened 2024-06-24
use tinybench for benchmarking

tokkiumai:


test/benchmark.js
```diff
@@ -70,14 +70,19 @@ suite
...
+  let hz = formatNumber(event.task.result.hz.toFixed(0)).padStart(10)
+  if (event.task.name === 'uid') {
+    name = '\nNon-secure:\n' + name
+  }
+  process.stdout.write(
+    `${name}${styleText('bold', hz)}${styleText('dim', ' ops/sec')}\n`
+  )
+})
+
+bench.threshold = 10
```
ai, line 85: What is the reason behind these changes?

tokkiumai: faster execution time. Threshold could be increased to a number of benchmark tasks (`bench._tasks.size`) but 10 is recommended by tinybench docs

ai: Can you explain why, can’t understand by docs what is default and what exactly we will change here? CI is almost identical.

tokkiumai: By default, benchmark tasks run synchronously. On nanoid benchmark suite the concurrent run not saving much time (it's -1 second for me: 21sec for synchronous run and 20 for concurrent). 

So, using concurrent warmup and run for benchmark is not something of a real need, since it gives about 4% better execution time

ai: Let’s close it since I'm afraid of some benchmark result issues with concurrent

PR #490, opened 2024-08-14
Fix CJS compatibility

FurryR:
# Proposed Changes

Ensure compatibility with some CDNs & bundlers on some CJS libraries.

jsDelivr translates `const { nanoid } = require('nanoid')` directly into `import e from 'nanoid/+esm'; const { nanoid } = e`, which is problematic and causes some libraries fail to load.

Here is an example using `postcss`:

```js
await import('https://cdn.jsdelivr.net/npm/postcss@8.4.41/+esm')
```

# Additional Information

It is suggested to squash merge this PR.

### DIFF
--- a/index.browser.js
+++ b/index.browser.js
@@ -61,4 +61,4 @@ export let nanoid = (size = 21) => {
   return id
 }
 
-export default { random, customRandom, customAlphabet, nanoid }
+export default { customAlphabet, customRandom, nanoid, random }
--- a/index.browser.js
+++ b/index.browser.js
@@ -60,3 +60,5 @@ export let nanoid = (size = 21) => {
   }
   return id
 }
+
+export default { random, customRandom, customAlphabet, nanoid }
--- a/index.d.ts
+++ b/index.d.ts
@@ -89,3 +89,5 @@ export const urlAlphabet: string
  * @returns An array of random bytes.
  */
 export function random(bytes: number): Uint8Array
+
+export default { customAlphabet, customRandom, nanoid, random }
--- a/index.js
+++ b/index.js
@@ -85,4 +85,4 @@ export function nanoid(size = 21) {
   return id
 }
 
-export default { random, customRandom, customAlphabet, nanoid }
+export default { customAlphabet, customRandom, nanoid, random }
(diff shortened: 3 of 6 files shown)


ai:
1. We also have [url-alphabet](https://github.com/ai/nanoid/tree/main/url-alphabet)
2. Please run `eslint --fix .`
3. I really worry that it could break something else. Can you test that this method doesn’t break Vite and webpack (you can add version from your branch by `npm install --save-dev nanoid@FurryR/nanoid`

FurryR:
> 1. We also have [url-alphabet](https://github.com/ai/nanoid/tree/main/url-alphabet)
> 2. Please run `eslint --fix .`
> 3. I really worry that it could break something else. Can you test that this method doesn’t break Vite and webpack (you can add version from your branch by `npm install --save-dev nanoid@FurryR/nanoid`

1. & 2. okay I will do that later
3. It seemed that adding a default export is not a breaking change at all, but I am unsure if the type description files (`.d.ts`) need to be updated. Anyway, I will take a look.

FurryR:
Tested with `pnpm start`. This PR does not break Vite.

ai:
> const { nanoid } = require('nanoid')

But Nano ID 4.x and 5.x don’t support `require()`. If you are using CJS you need 3.x version. We have a special branch for it.

ferreiro:
@ai any way you can guys ship this? I was trying using nanoId with typescript the problem is that when I transpile and push the code to google app engine, it says that the require() is not supported. Maybe you have a workaround instead of needing this PR?

ai:
@ferreiro as I told I need some investigation that it will not break in many possible cases where Nano ID is used.

FurryR:
> > const { nanoid } = require('nanoid')
> 
> But Nano ID 4.x and 5.x don’t support `require()`. If you are using CJS you need 3.x version. We have a special branch for it.

Hmm. weird. If you access [https://cdn.jsdelivr.net/npm/nanoid@3.3.7/non-secure/+esm](https://cdn.jsdelivr.net/npm/nanoid@3.3.7/non-secure/+esm) you will see.

FurryR:
Anyway, I will close this PR and wait for a better workaround. Thank you all guys.

PR #504, opened 2024-10-27
use `Math.log2` instead of `Math.log / Math.LN2`

kirillgroshkov:
I have noticed that this line in `index.browser.js`:

```js
let mask = (2 << (Math.log(alphabet.length - 1) / Math.LN2)) - 1
```

uses `Math.log() / Math.LN2` to get the `log2`,

while it can be made shorter by using widely available `Math.log2` method:

```js
let mask = (2 << Math.log2(alphabet.length - 1)) - 1
```

How did I notice it?

Because there's an eslint rule that recommends that: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-modern-math-apis.md#prefer-mathlog2x-over

`Math.log2` seems to be widely available in the Browsers and I see no reason why it shouldn't be used.


ai:
Thanks. Released in 5.0.8.

Issue #347, opened 2022-02-10
Underscore is sometimes an unsafe character


stefansundin:
Hello there,

Just wanted to open this issue to discuss the underscore character. It is unsafe if used at the very end of an URL on GitHub. (I also tested it on Twitter but they handle it correctly.)

Here's an example: https://example.com/abcdefgh_

I noticed this problem because I use https://www.npmjs.com/package/react-markdown in a project and it rendered a link in this broken way which did ultimately cause problems for a user. If it is happening to me then I'm sure other people are experiencing it too (perhaps unwittingly since it is kinda rare, 1/64 chance).

If a non-`_` character is added at the end then the link is no longer broken. https://example.com/abcdefgh_/

Anyway, I wanted to start a discussion here to raise the problem. Perhaps the react-markdown package can be improved to handle this situation? Are there other libraries and packages that may misbehave in a similar way?

I am going to use a custom alphabet in my project. Although I find that the API for using a custom alphabet is a bit cumbersome. I will open a PR soon with a suggested API change (but I need help to finish it).

Thanks!

ai:
Interesting case.

What symbol can we use in return to keep the same size of default alphabet?

ehoogeveen-medweb:
I don't think there are any perfect options - the only characters that are unreserved in all contexts for URIs (aside from letters and numbers) are `-`, `.`, `_` and `~` - and of those, only `-` stays as part of the link on GitHub (and `-` is already part of the standard alphabet).

I think the closest would be `$` - it's reserved as a sub-delimiter and correspondingly `encodeURIComponent` will encode it, but it doesn't have any specified special meaning and it's valid for use in path segments (see e.g. section 3.3 of [RFC 3986](http://www.faqs.org/rfcs/rfc3986.html)).

Example: https://example.com/abcdefgh$

linhub15:
Wrap the link with angle brackets `<` `>` to render the URL correctly in Github and probably most other Markdown libraries.
* `<https://example.com/abcdefgh_>` correctly renders to <https://example.com/abcdefgh_>

Since Markdown uses underscores to denotes italic text, the parser cannot accurately decide if `https://a_` is a link. Markdown parsers are pretty good at guessing that `https://a` is a link most of the time, so we're conditioned to writing links without the angle brackets.

Underscores are definitely URL safe.
They are included in the list of special characters that can be used without encoding [(RFC 1738, Section 2.2)](https://www.ietf.org/rfc/rfc1738.html#section-2.2).
[...]

PR #514, opened 2025-01-01
solution to implement url-safe strings (#347)

Mindslayer001:
### Description:
This pull request introduces an approach to generate URL-safe strings, as discussed in Issue #347. It implements a new `--url-safe` (or `-u`) flag for the `nanoid` CLI, enabling users to generate IDs that are safe to use in URLs by avoiding problematic characters, such as underscores at the end of strings.

### Key Changes:
- **New URL-Safe Alphabet**: A custom alphabet is now used for generating URL-safe IDs. It excludes characters like underscores and other potentially problematic characters in URLs.
- **CLI Update**: The `--url-safe` flag has been added to the command-line interface, allowing users to specify that the generated ID should be URL-safe.
- **Enhanced `customAlphabet` Function**: The `customAlphabet` function has been updated to accept a `url_safe` parameter, making it easy for users to generate URL-safe IDs when needed.

### Motivation:
This approach addresses issues related to trailing underscores and other characters that can cause problems in URLs, such as rendering issues or broken links in Markdown and web applications. By enabling the generation of URL-safe IDs, the solution enhances compatibility and reliability when working with URLs.

### Key Note:
This is not yet considered a complete solution, but rather an approach to how this can be implemented. Feedback is welcome for further improvement.

### DIFF
--- a/bin/nanoid.js
+++ b/bin/nanoid.js
@@ -19,17 +19,20 @@ Options
   -s, --size       Generated ID size
   -a, --alphabet   Alphabet to use
   -h, --help       Show this help
+  -u, --url-safe   Generate url-safe string
 
 Examples
   $ nanoid -s 15
   S9sBF77U6sDB8Yg
 
   $ nanoid --size 10 --alphabet abc
+  bcabababca)
+  $ nanoid --size 5 -u 1
   bcabababca`)
   process.exit()
 }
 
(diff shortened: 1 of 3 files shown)


ai:
Sorry, I don’t plan to change alphabet. I think #347 can be better solved by adding `/` to the end of URL.

Issue #468, opened 2024-02-20
Not working with expo 


2ico:
Expo SDK: 49.0.0
After following the instructions at https://github.com/ai/nanoid#react-native , I get:
```
Android Bundling failed 403ms
The package at "node_modules/nanoid/index.js" attempted to import the Node standard library module "node:crypto".
It failed because the native React runtime does not include the Node standard library.
```

ai:
Did you foolow that steps?

https://github.com/ai/nanoid?tab=readme-ov-file#react-native

2ico:
Yes, I did

ai:
Sorry, I have no idea or experience with Expo.

Try to ask Expo community.

2ico:
Async was removed in v5, right?

ai:
Yes

2ico:
Can I contact them?

ai:
I don’t think it is polite. You should use StackOverflow and similar things if you want ask others to do your work.

Try also import `index.browser.js` version. Maybe `react-native-get-random-values` polyfill has issue with import.

2ico:
I don't think it's impolite to contact the maintainer of a library.

ai:
I am the maintainer. You want to contact some other developer who send PR to the library.

2ico:
Anyway, how can I do this?

> Try also import `index.browser.js` version. Maybe `react-native-get-random-values` polyfill has issue with import.

```
Could not find a declaration file for module 'nanoid/index.browser'. '/home/george/Documents/GAIO/expo/archie-expo/node_modules/nanoid/index.browser.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/nanoid` if it exists or add a new declaration (.d.ts) file containing `declare module 'nanoid/index.browser';`
  ```

**Comment by ai** [comment]
This is just a type warning not an error.

Does it works?

**Comment by ai** [comment]
Try the new Nano ID 5.0.6 with normal `import { nanoid } from 'nanoid`

**Comment by 2ico** [comment]
It doesn't work, unfortunately. It's still using './index.js'

**Comment by 2ico** [comment]
And I guess the release should include a ./index.browser.d.ts file to solve the error I got with TS.

**Comment by steida** [comment]
I tried this patch but without a success

```
diff --git a/package.json b/package.json
index 0d10ada4dce6c02a339399e71c82860c70f3d12b..4fb2ab8de9093eae1bc590cd738019e5b40d0dfb 100644
[...]

PR #515, opened 2025-01-05
Fix Expo, #468

steida:


ai:
1. Can you give me some proofs that React Native support `global.crypto` right now
2. Do we need to change docs? https://github.com/ai/nanoid?tab=readme-ov-file#react-native

steida:
I use only this:

```ts
import { install } from "react-native-quick-crypto";
install();
```

The same fix works for both NanoId versions (3/5).

Here is my package.json:

```json
"type": "module",
  "types": "./dist/src/index.d.ts",
  "exports": {
    ".": {
      "types": "./dist/src/index.d.ts",
      "import": "./dist/src/index.js",
      "browser": "./dist/src/index.js",
      "react-native": "./dist/src/index.js"
    },
    "./evolu": {
      "import": "./dist/src/Evolu/Internal.js"
    }
  },
  "typesVersions": {
    "*": {
      "evolu": [
        "./dist/src/Evolu/Internal.d.ts"
      ]
    }
  },
```

Btw, I think it should be mentioned that it's recommended to use nanoid 3 in monorepos with hoisting enabled because many packages are still using nanoid 3 (which is not dead; that should also be mentioned). Anyway, thank you for your work.

ai:
Honestly, this PR looks like quick fix which will break workflow of many users because the changes was not careful considered and planned.

> import { install } from "react-native-quick-crypto";
install()

Is the code will fail without it?

Why we are not mentioning it in the docs?

Why it is better than previous polyfill?

> Btw, I think it should be mentioned that it's recommended to use nanoid 3 in monorepos with hoisting enabled because many packages are still using nanoid 3 

We can release the be version of 3.x with the new polyfill

steida:
All I know is that it allows me to use NanoId in Expo/RN without hacks.

PR #521, opened 2025-02-14
Make functions generic so it can cast to opaque (branded) types

kossnocorp:
The PR introduces generics to the Nano ID functions that return `string`.

Right now, when using opaque (branded) strings, you have cast explicitly, which opens room for an error:

```ts
export function newWorkbenchDatasetSelect(): Workbench.CellDatasetSelect {
  return {
    type: "select",
    id: nanoid(),
    //=> Type 'string' is not assignable to type 'CellDatasetId'.
    datasetUid: undefined,
    selection: newDatasetSelectionSolo(),
  };
}
```

Before the only option was to use `as`:

```ts
export function newWorkbenchDatasetSelect(): Workbench.CellDatasetSelect {
  return {
    type: "select",
    id: nanoid() as Workbench.CellDatasetId,
    datasetUid: undefined,
    selection: newDatasetSelectionSolo(),
  };
}
```

After the change, in most cases, the code won't require casting at all, as TypeScript will infer it automatically.

In cases where it is not possible, the Nano ID users will have an option to specify the type parameter:

```ts
mockWorkbenchDatasetSelect(nanoid<Workbench.CellDatasetId>());
```

It will not break anything for existing users, and it doesn't impact the build size, so there are no downsides (screenshot after applying the patch):

image

### DIFF
--- a/README.id-ID.md
+++ b/README.id-ID.md
@@ -302,6 +302,29 @@ LZfXLFzPPR4NNrgjlWDxn
 
 Bila ingin mengganti alfabet atau ukuran ID, dapat menggunakan [`nanoid-cli`](https://github.com/twhitbeck/nanoid-cli).
 
+### TypeScript
+
+Nano ID memungkinkan untuk mengubah string yang dihasilkan menjadi string opak
+dalam TypeScript. Sebagai contoh:
+
+```ts
+type UserId = string & { [userIdBrand]: true }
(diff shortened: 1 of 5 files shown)


ai:
Nice feature! Thanks.

ai:
Released in 5.1.

kossnocorp:
Thank you a lot! 🚀

PR #577, opened 2026-04-12
Optimize customAlphabet rejected-byte handling

saripovdenis:
## What

Optimize `customAlphabet()` by skipping empty string appends for rejected random bytes in both the Node and browser implementations.

Also raise the `customAlphabet` size budget from `173 B` to `174 B`.

## Why

`customAlphabet()` is built on top of `customRandom()`. Today, when a masked random byte falls outside the alphabet range, the code still does an empty-string append through:

```js
id += alphabet[bytes[i] & mask] || ''
```

This change avoids that extra string work by rejecting the byte with `continue` before the append.

This is not a universal win for every possible alphabet. In particular, for alphabets with very low or zero rejection rates, the extra branch can be slightly slower. But for reject-heavy alphabets, it improves throughput by avoiding unnecessary string concatenations for discarded bytes.

That tradeoff matches the repo’s current `customAlphabet` benchmark case, `customAlphabet('1234567890abcdef-', 10)`, where rejected bytes are common. In paired local measurements of that exact case, median throughput improved from `5,782,044 ops/sec` to `6,197,715 ops/sec` (`+7.25%`), with the optimized version faster in `60/60` samples.

## Changes

```diff
- id += alphabet[bytes[i] & mask] || ''
+ if (!alphabet[bytes[i] & mask]) continue
+ id += alphabet[bytes[i] & mask]
```

Apply the same change in both:

- `index.js`
- `index.browser.js`

And update the package size budget:

```diff
- "limit": "173 B"
+ "limit": "174 B"
```

## Size Impact

`customAlphabet` increases by `1 B` in the bundled, minified, brotlied size check:

- before: `173 B`
- after: `174 B`

### DIFF
--- a/index.browser.js
+++ b/index.browser.js
@@ -41,8 +41,9 @@ export let customRandom = (alphabet, defaultSize, getRandom) => {
       // A compact alternative for `for (var i = 0; i < step; i++)`.
       let j = step | 0
       while (j--) {
-        // Adding `|| ''` refuses a random byte that exceeds the alphabet size.
-        id += alphabet[bytes[j] & mask] || ''
+        // Adding `continue` refuses a random byte that exceeds the alphabet size.
+        if (!alphabet[bytes[j] & mask]) continue
+        id += alphabet[bytes[j] & mask]
         if (id.length >= size) return id
       }
     }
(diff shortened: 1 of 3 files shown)


ai:
Nice addition, thanks

saripovdenis:
@ai i'm also working on some optimisation for customAlphabet, will show you today

might be worth of publishing

PR #582, opened 2026-04-12
Use full byte range in customAlphabet (+ ~75% ops / sec optimisation)

saripovdenis:
## What

Optimize `customAlphabet()` by replacing the old next-power-of-two mask + rejection scheme with an unbiased threshold over the full `0..255` byte range in both the Node and browser implementations.

Keep the bitmask fast path for power-of-two alphabets, add zero-size empty-alphabet coverage, update deterministic `customRandom()` expectations, and raise the `customAlphabet` size budget from `178 B` to `183 B`.

## Why

`customAlphabet()` is built on top of `customRandom()`. The previous implementation always rounded the alphabet length up to the next power of two, built a `mask`, and then rejected masked values that landed outside the alphabet range.

That works well for power-of-two alphabets, but it wastes a lot of bytes for awkward alphabet sizes, especially sizes just above a power of two. For example, with a 17-character alphabet, the old code masked into `0..31` and could only use `17` of those `32` outcomes. The new logic can use `255` of `256` byte values.

This change computes `max = 256 - (256 % alphabet.length)`, accepts only bytes below `max`, and maps accepted bytes with `% alphabet.length`. When `max === 256`, it keeps the bitmask fast path.

The biggest wins are for reject-heavy alphabet sizes. Accepted-byte rate examples:

- `17` chars: `53.1%` -> `99.6%`
- `21` chars: `65.6%` -> `98.4%`
- `33` chars: `51.6%` -> `90.2%`
- `70` chars: `54.7%` -> `82.0%`

This is not a universal win for every possible alphabet. Power-of-two sizes and already-efficient sizes such as `16`, `31`, `32`, `63`, `64`, `100`, `129`, and `200` stay roughly the same.

In a local `tinybench` run on this machine for the repo’s benchmark case, `customAlphabet('1234567890abcdef-', 10)`, throughput improved from `5,237,968 ops/sec` to `9,103,488 ops/sec` (`+73.80%`).

## Changes

```diff
- let mask = ...
- let step = Math.ceil((1.6 * mask * defaultSize) / alphabet.length)
+ let max = 256 - (256 % alphabet.length)
+ let step = Math.ceil((1.6 * 256 * defaultSize) / max)

...

- let next = alphabet[bytes[i] & mask]
- if (!next) continue
- id += next
+ if (max === 256) {
+   id += alphabet[bytes[i] & (alphabet.length - 1)]
+ } else if (bytes[i] < max) {
+   id += alphabet[bytes[i] % alphabet.length]
+ }
```

Apply the same logic in both:

- `index.js`
- `index.browser.js`

Also:

- add zero-size empty-alphabet coverage in `test/index.test.js`
- update the deterministic `customRandom()` expectations
- raise the package size budget:

```diff
- "limit": "178 B"
+ "limit": "193 B"
```

## Size Impact

`customAlphabet` increases by `15 B` in the bundled, minified, brotlied size check:

- before: `178 B`
- after: `193 B`

### DIFF
--- a/index.browser.js
+++ b/index.browser.js
@@ -10,28 +10,13 @@ export { urlAlphabet } from './url-alphabet/index.js'
 export let random = bytes => crypto.getRandomValues(new Uint8Array(bytes))
 
 export let customRandom = (alphabet, defaultSize, getRandom) => {
-  // First, a bitmask is necessary to generate the ID. The bitmask makes bytes
-  // values closer to the alphabet size. The bitmask calculates the closest
-  // `2^31 - 1` number, which exceeds the alphabet size.
-  // For example, the bitmask for the alphabet size 30 is 31 (00011111).
-  // `Math.clz32` is not used, because it is not available in browsers.
-  let mask = (2 << Math.log2(alphabet.length - 1)) - 1
-  // Though, the bitmask solution is not perfect since the bytes exceeding
-  // the alphabet size are refused. Therefore, to reliably generate the ID,
-  // the random bytes redundancy has to be satisfied.
-
-  // Note: every hardware random generator call is performance expensive,
-  // because the system call for entropy collection takes a lot of time.
-  // So, to avoid additional system calls, extra bytes are requested in advance.
-
-  // Next, a step determines how many random bytes to generate.
-  // The number of random bytes gets decided upon the ID size, mask,
-  // alphabet size, and magic number 1.6 (using 1.6 peaks at performance
-  // according to benchmarks).
-
-  // `-~f => Math.ceil(f)` if f is a float
-  // `-~i => i + 1` if i is an integer
-  let step = -~((1.6 * mask * defaultSize) / alphabet.length)
+  // `max` is the largest unbiased slice of the 0-255 byte range.
+  // If the alphabet size divides 256, every byte is usable.
+  // This usually rejects fewer bytes than the original bitmask approach.
+  let max = 256 - (256 % alphabet.length)
+  // Use a fixed batch size based on the unbiased byte range.
+  // `1.6` is a magic number chosen from benchmarks.
+  let step = Math.ceil((1.6 * 256 * defaultSize) / max)
 
   return (size = defaultSize) => {
     if (!size) return ''
@@ -41,10 +26,11 @@ export let customRandom = (alphabet, defaultSize, getRandom) => {
       // A compact alternative for `for (var i = 0; i < step; i++)`.
       let j = step
       while (j--) {
-        let next = alphabet[bytes[j] & mask]
-        // Adding `continue` refuses a random byte that exceeds the alphabet size.
-        if (!next) continue
-        id += next
+        if (max === 256) {
+          id += alphabet[bytes[j] & (alphabet.length - 1)]
+        } else if (bytes[j] < max) {
+          id += alphabet[bytes[j] % alphabet.length]
+        }
         if (id.length >= size) return id
       }
     }
--- a/index.browser.js
+++ b/index.browser.js
@@ -10,15 +10,26 @@ export { urlAlphabet } from './url-alphabet/index.js'
 export let random = bytes => crypto.getRandomValues(new Uint8Array(bytes))
 
 export let customRandom = (alphabet, defaultSize, getRandom) => {
-  // `max` is the largest unbiased slice of the 0-255 byte range.
-  // If the alphabet size divides 256, every byte is usable.
-  // This usually rejects fewer bytes than the original bitmask approach.
-  let max = 256 - (256 % alphabet.length)
-  // Use a fixed batch size based on the unbiased byte range.
+  // Random bytes are 0-255 and already have full entropy.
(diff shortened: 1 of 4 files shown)


saripovdenis:
@ai not sure "1.6" is still magical, but for current benchmark still makes it faster

saripovdenis:
@ai Do you need some extra explanation about this PR? I'm ready to provide details

ai:
Sorry, I will first review other PRs in my open source projects (`nanostores/i18n` and `ai/size-limit` is above in the queue).

I hope to review this PR on this week.

saripovdenis:
> Sorry, I will first review other PRs in my open source projects (`nanostores/i18n` and `ai/size-limit` is above in the queue).
> 
> I hope to review this PR on this week.

Got it! Thanks!

index.js
```diff
@@ -58,10 +47,12 @@ export function customRandom(alphabet, defaultSize, getRandom) {
       // A compact alternative for `for (let i = 0; i < step; i++)`.
       let i = step
       while (i--) {
-        let next = alphabet[bytes[i] & mask]
-        // Adding `continue` refuses a random byte that exceeds the alphabet size.
-        if (!next) continue
-        id += next
+        if (max === 256) {
```
ai: `max` is constant. You do not need to call `if` on every symbol. Just return different functions.

Don't afraid of code duplication here.

index.browser.js
```diff
@@ -41,10 +26,12 @@ export let customRandom = (alphabet, defaultSize, getRandom) => {
...
       while (j--) {
-        let next = alphabet[bytes[j] & mask]
-        // Adding `continue` refuses a random byte that exceeds the alphabet size.
-        if (!next) continue
-        id += next
+        if (max === 256) {
+          // For power-of-two alphabets, this bitmask is the same as modulo, but faster.
+          id += alphabet[bytes[j] & (alphabet.length - 1)]
+        } else if (bytes[j] < max) {
+          id += alphabet[bytes[j] % alphabet.length]
```
ai, line 62: `random % alphabet` is an unsafe way to generate symbols because the distribution of each symbol will be different (and secret ID will be easier to guess).

We are using `random % alphabet` in our demo page (run it by `pnpm start`) it will generate you an example why it doesn’t work.

But in short: Example: if you map random numbers `0...9` into `ABC` alphabet with `% 3`, you get:

```
0,3,6,9 → A
1,4,7 → B
2,5,8 → C
```

So `A` appears more often than `B` or `C`.

saripovdenis: yes, but the approach here:
```
max = 10 - (10 % 3) // max = 9
// if generated byte >= 9 -> reject
// so the range is 0,1...8
// each character has equal chances to be rolled
0,3,6 → A
1,4,7 → B
2,5,8 → C
```

Another example with high rejection rate:
```
alphabet.length = 17

old approach:
- find closest bitmask bigger than alphabet.length
- reject everything after alphabet.length

mask = 31 (closest power of two bigger than 17 minus 1)
accepted range: 0,...,16
rejected range: 17,...31 // so rejections were very likely

new approach:
- find the max byte size so each character has equal chance to be rolled
- reject bytes that gives unfair chance for some of the characters

max = 255 // 256 - (256 % 17)
accepted range: 0,...254 // each character has 15 times bigger chance to be rolled. But still chances for each character are equal
rejected range: 255
```

in the current solution we just increase chance of each character to be randomised fairly (the same chance for all characters), but decreasing the rejection range from whats left between alphabet.length and bitmask to modulo from 256 and alphabet length

_256 was choosen as largest secure supported alphabet length_

P.s: hope this explains the idea well haha, but I'm also would like to ensure entropy stays the same as previous approach

ai: I see. Let's change variable names (`max` is too confusing) to make it easier to do security review for users.

Also we need to document it.

index.js
saripovdenis: Done!

index.browser.js
```diff
@@ -10,28 +10,32 @@ export { urlAlphabet } from './url-alphabet/index.js'
...
-  let step = -~((1.6 * mask * defaultSize) / alphabet.length)
+    return (size = defaultSize) => {
+      if (!size) return ''
+      let id = ''
+      while (true) {
+        let bytes = getRandom(step)
+        // A compact alternative for `for (var i = 0; i < step; i++)`.
+        let j = step
+        while (j--) {
+          // For power-of-two alphabets, this bitmask is the same as modulo, but faster.
```
ai: We lost explanation why we are using mask here

saripovdenis: i will finish this PR later today

saripovdenis: @ai Added explanation to avoid concerns from users. Let me know if there is something else would be good to change in this PR!

index.browser.js
saripovdenis: > I see. Let's change variable names (`max` is too confusing) to make it easier to do security review for users.
> 
> Also we need to document it.

@ai I choose "safeByteCutoff" - hope it translates the meaning well

ai:
Thanks!

ai:
Released in 5.1.8.

PR #594, opened 2026-05-25
Exclude local config from npm package

### LINKED ISSUE #590: Published tarball includes `.claude/settings.local.json` (add `files` allowlist or `.npmignore`)
## Summary

v3.3.12 (the version currently installed in my project) ships `.claude/settings.local.json` inside the npm tarball because `package.json` has no `files` allowlist and the repo has no `.npmignore`.

## Repro

```
npm pack nanoid@latest
tar tzvf nanoid-*.tgz | grep -E '\.claude|CLAUDE\.md'
```

## What's leaking

`.claude/settings.local.json` — a Claude Code permissions file with paths like `/workspaces/nanoid/` (suggesting a Codespaces dev container). No runtime purpose for consumers; pure maintainer-local workflow artifact.

## Why this matters

Files in `.claude/` are intended for the developer's Claude Code session. Shipping them to every install:
- Adds bytes to the tarball (small here, but the pattern matters)
- Exposes maintainer dev paths in the published artifact
- Can be auto-loaded by downstream developers' Claude Code sessions if they ever `cd` into `node_modules/nanoid/`

## Suggested fix

Add a `files` allowlist to `package.json` (preferred — allowlist beats denylist for npm publishing):

```json
"files": ["index.js", "index.d.ts", "nanoid.js", "url-alphabet/", ...]
```

Or add `.claude` to `.npmignore`.

## Context

Found during a Claude Code permissions audit on my machine. Several other npm packages have the same pattern (`resolve`, `sitemap`, a few MCP servers); filing individually so each maintainer can decide on a fix path. Happy to PR if useful.

mturac:
## Summary

- Exclude maintainer-local configuration files from the npm package
- Keep the current package ignore approach and avoid changing runtime exports

## Verification

- Added temporary local config files and ran `npm pack --dry-run --json`; confirmed none were included in the packed file list
- Ran `npm pack --dry-run --json` again after removing the temporary files
- Ran `git diff --check`

Fixes #590

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,6 +2,9 @@ test/
 tsconfig.json
 coverage/
 
+.claude/
+CLAUDE.md
+
 img/
 
 jsr.json


ai:
We use `clean-publish` to fix it.

We have `.claude` not in the latest 5.x, but in old 3.x. Fixing it is much harder because of how npm works.

I am planning to do it on next week.

PR #600, opened 2026-06-23
fix(non-secure): clamp a negative size instead of looping forever

spokodev:
The secure module already guards against an out-of-range `size`: `fillPool` throws a `RangeError` and the generators short-circuit empty input. The `nanoid/non-secure` twin was missed.

Both `nanoid` and `customAlphabet` in `non-secure/index.js` count down with `let i = size | 0; while (i--)`. `size | 0` truncates fractional sizes but does not clamp negatives, so a negative `size` decrements `i` toward `-2^31` without ever reaching `0`. The string keeps growing, the synchronous loop blocks the event loop, and the process hangs until it runs out of heap (a denial-of-service trigger):

```js
import { nanoid, customAlphabet } from 'nanoid/non-secure'

nanoid(-1)                 // hangs forever -> OOM
customAlphabet('abc')(-1)  // hangs forever -> OOM
```

#508 and #510 established size-driven infinite loops as an in-scope bug class, but #510's non-secure change (`size | 0`) only addressed the fractional case, not negatives.

## Fix

Clamp the counter with `Math.max(size | 0, 0)` in both functions, so a negative size yields `''` (the same result as `nanoid(0)`). One extra function call, no behavior change for valid sizes.

## Verification

With the fix reverted, the new negative-size test hangs: `node --test` produces no output and is killed by a hard 6s cap (exit 142, SIGALRM) because the blocked loop never lets the reporter flush. With the fix applied:

- `nanoid(-1) === ''`, `nanoid(-100) === ''`
- `customAlphabet('abcdef')(-1) === ''`, `customAlphabet('abcdef', -5)() === ''`
- Regressions preserved: `nanoid(0) === ''`, `nanoid(2.9).length === 2`, `nanoid(10).length === 10`
- Full suite green (non-secure, index, bin): 65 tests pass, 0 fail

### DIFF
--- a/non-secure/index.js
+++ b/non-secure/index.js
@@ -13,8 +13,9 @@ export let customAlphabet = (alphabet, defaultSize = 21) => {
   return (size = defaultSize) => {
     let id = ''
     // A compact alternative for `for (var i = 0; i < step; i++)`.
+    // `> 0` stops a negative size from looping forever.
     let i = size | 0
-    while (i--) {
+    while (i-- > 0) {
       // `| 0` is more compact and faster than `Math.floor()`.
       id += alphabet[(Math.random() * alphabet.length) | 0]
     }
@@ -25,8 +26,9 @@ export let customAlphabet = (alphabet, defaultSize = 21) => {
 export let nanoid = (size = 21) => {
   let id = ''
   // A compact alternative for `for (var i = 0; i < step; i++)`.
+  // `> 0` stops a negative size from looping forever.
   let i = size | 0
-  while (i--) {
+  while (i-- > 0) {
     // `| 0` is more compact and faster than `Math.floor()`.
     id += urlAlphabet[(Math.random() * 64) | 0]
   }
--- a/package.json
+++ b/package.json
@@ -97,13 +97,13 @@
       "name": "non-secure nanoid",
       "import": "{ nanoid }",
       "path": "non-secure/index.js",
-      "limit": "90 B"
+      "limit": "93 B"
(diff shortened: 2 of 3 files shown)


ai:
30% size increase (90 → 120 B) is too much for fixing only theoretical issue.

spokodev:
You are right that +30B was too much, but that cost was the `Math.max()` call, not the fix itself. I switched to `while (i-- > 0)` (keeping `i = size | 0`), which stops the negative-size loop for +3B on `nanoid` and +2B on `customAlphabet`, measured with size-limit:

| | nanoid | customAlphabet |
|---|---|---|
| main | 90 B | 53 B |
| `Math.max` (before) | 120 B | 66 B |
| `while (i-- > 0)` | 93 B | 55 B |

A zero-byte fix is not possible: `size | 0` has to stay (it keeps `nanoid(2.9) === 2`), and once it stays, stopping a negative loop costs at least one byte. I bumped the two limits to 93/55 accordingly.

On whether it is worth it: the secure `nanoid` already throws on `bytes < 0` and the CLI rejects `size <= 0`, so the non-secure twin is the only entry point that hangs instead of failing fast. The existing `size | 0` guard handles the fractional case but not a negative size, which is a synchronous infinite loop that grows the string until the process runs out of memory. I kept clamp-to-empty (matches `nanoid(0)`); if you would prefer it to throw like the secure module, that is a few bytes more and I can switch.

ai:
Thanks. Released in 5.1.16.

PR #601, opened 2026-06-30
fix(non-secure): clamp negative size to prevent infinite loop

mahirhir:
On the `v3` branch the non-secure `nanoid` and `customAlphabet` still use `while (i--)`, which never terminates for a negative `size`. `nanoid(-1)` keeps appending to the string and grows until the process runs out of memory.

This was already fixed on main in #600 (released in 5.1.16) by changing the guard to `while (i-- > 0)`, but the `v3` branch (latest 3.3.15) didn't get it.

This backports the same guard to both `non-secure/index.js` and `non-secure/index.cjs`, and adds the negative-size tests from #600 in v3's uvu style.

For non-negative sizes the loop runs the exact same number of iterations as before, so there is no behavior change for normal input.

Test plan: `npx uvu . "non-secure.test.cjs$"` passes 11/11. Without the guard the two new negative-size tests hang the runner; with it `nanoid(-1)` and `customAlphabet('abcdef')(-1)` return an empty string.

### DIFF
--- a/non-secure/index.cjs
+++ b/non-secure/index.cjs
@@ -11,8 +11,9 @@ let customAlphabet = (alphabet, defaultSize = 21) => {
   return (size = defaultSize) => {
     let id = ''
     // A compact alternative for `for (var i = 0; i < step; i++)`.
+    // `> 0` stops a negative size from looping forever.
     let i = size | 0
-    while (i--) {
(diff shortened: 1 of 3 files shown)


ai:
Thanks. Released in 3.3.16.

PR #609, opened 2026-09-01
Reject alphabets the generator cannot sample from

lenamonj:
`customAlphabet` and `customRandom` accept alphabets they cannot sample from, and the failure is a hang rather than an error:

```js
customAlphabet('')();              // spins forever in while (true), both secure builds
customAlphabet('a'.repeat(300))(); // same hang on the node build
```

The non-secure build returns IDs made of the literal text `undefined` for an empty alphabet. Reachable from the CLI too: `nanoid --alphabet "$AN_EMPTY_OR_LONG_VALUE"` hangs with no output, no error, exit only by kill. The test suite has full line, branch and function coverage - every one of these sits behind a case that only ever calls the generator with size 0, which is why coverage never saw it.

**Fix**: each module validates the alphabet once, at the boundary the argument enters, and rejects at construction time with `RangeError: Wrong alphabet size` (the message style the size check already uses). The bounds follow each implementation's real contract: the secure builds require 1..256 symbols, because a random byte addresses the alphabet; the non-secure build requires only non-empty, because it indexes rather than byte-addresses and genuinely supports longer alphabets - the tests pin the positive side of each bound (a 256-symbol alphabet still generates on the secure builds, a 300-symbol one on the non-secure build), so a guard that merely rejected more would fail them.

**Size**: the guard grows two `size-limit` budgets - `customAlphabet` and non-secure `customAlphabet` - and those two entries are moved to their measured sizes in this PR. `nanoid` itself, the headline number, is untouched.

**Verification**: the new tests fail on `main` and pass with the fix; the full `pnpm test` (oxlint, tsc, node --test with coverage, size-limit) is green. `index.d.ts` and the README's custom-alphabet sentence are updated to state the contract.

---

Provenance, for transparency: this defect was found by an automated audit loop I run against open-source projects; the patch and its verification were reviewed by me before filing, and I'm happy to answer any questions during review.


package.json
```diff
@@ -86,7 +86,7 @@
     {
       "name": "customAlphabet",
       "import": "{ customAlphabet }",
-      "limit": "207 B"
+      "limit": "239 B"
```
ai: It is too big price for the case which we haven't in reality.

Let's just add notice to docs.

lenamonj:
Agreed - reworked to a docs-only notice. The alphabet sentence in the custom-alphabet section now reads "from 1 to 256 symbols" and warns that a generator built outside those bounds can loop forever. Code, types, tests and the size budgets are back to main; the diff is now README-only.

ai:
Thanks!

PR #610, opened 2026-09-22
docs: fix custom generator JSDoc examples

Likio3000:
Two JSDoc examples in `index.d.ts` need adjustments: the `urlAlphabet` example calls `customAlphabet` without importing it, and the `customRandom` example returns a plain array even though the callback's declared return type is `Uint8Array`. Import the referenced factory and construct a typed byte array in the callback.

Only comments change; the type declarations and runtime behavior are unchanged. Checked the examples against the exports and callback signature, confirmed the declarations are identical after removing comments, and syntax-checked the JavaScript fences without executing them. TypeScript compilation and the project suite were not run. Prepared and reviewed with AI assistance.


ai:
Thanks!

